PR #160, opened 2016-01-08
Query authenticator information at the SOAP interface

RKrahl:
The last icat.server snapshot adds discovery of available authenticators and the credential keys they expect through the new `properties` call at the RESTful interface.  This information would be also useful to SOAP clients, thus I submit this pull request that adds a corresponding call to the SOAP interface.

There are some issues that may be worth a discussion:

1. The name `getProperties` is already taken at the SOAP interface for a different call.  For this reason, I named the new call `getAuthenticatorInfo`.

2. I restricted the scope of the call to just the authenticator information.  The new `properties` call at the RESTful interface also returns maxEntities and lifetimeMinutes, but I didn't consider these very important, so I left them out.  On the other hand, it might be desirable to keep both interfaces consistent as far as possible.

3. Most of the work implementing this call was put into decoding the Json string that the authenticators return in `getDescription()`.  This looks a bit silly that the authenticators first encode this information as a Json string and then another class in the same application needs to decode this string again to get the information back.

  An alternate implementation would be to move the new `AuthenticatorDescription` and `AuthenticatorCredentialKey` classes to icat.authentication (renaming them to `Description` and `CredentialKey` respectively) and change the `Authenticator.getDescription()` call to return a `Description` rather then a Json string.  This `Description` should then get a `toJsonObject()` method to be called from `ICATRest.getProperties()`.  This alternate implementation would seem cleaner and leaner to me.  But it would require changing icat.authentication and the authenticators as well.


RKrahl:
I hit submit too ealy, two further remarks:

1. I tested this code with the sources of the last icat.server snapshot release.  I can't compile the latest icat.server source version from the git repository here, as it requires icat.utils 4.12.0-SNAPSHOT that is not found in the maven repository.  But I don't believe that this makes any difference.

2. Client support for the new `getAuthenticatorInfo` call is also available in the authenticatorInfo branch in python-icat.  This also demonstrates how this call could be used from the client side and how a client may adapt itself to the authenticator info from the server.  E.g. I tested this to actually work with client and server.

fisherab:
I don't agree with this change for a number of reasons. The main one being that I don't want to add code to complicate the old SOAP API. For python-icat the most obvious thing to do is to access both the SOAP and rest APIs and use whichever is most appropriate for your call. You already access the IDS using a rest API.  The maxEntities property is needed by the IDS so that it does not produce calls to ICAT that will fail and the same is true for anybody else using ICAT. I agree that lifetimeMinutes is of less value as you can log in and then see how much time you have left however I prefer to have direct access to the property. One of the benefits of the restful API is that the set of properties being returned can easily be modified without breaking existing code provided that this is done in a reasonable manner.

RKrahl:
> I don't agree with this change for a number of reasons.  The main
> one being that I don't want to add code to complicate the old SOAP
> API.

First of all, I firmly disagree that SOAP is old and RESTful is the way to go.  SOAP has many advantages over RESTful that makes life much easier on the client side.  Most importantly, SOAP is self documenting.  You only need to have a look at the WSDL to see exactly the structure of the messages that the server expects and delivers, including all the data types.  This WSDL is machine readable and that means the the client can automatically check whether it is compatible with this server's version.

We are now at ICAT 4.5.1 and python-icat still supports all ICAT versions since 4.2.* (aside from some newer features that require JPQL queries).  It mends some of the incompatibilities automatically, such that client code doesn't even see them.  Such a tolerance of diverting server versions would be extremly difficult to maintain in a RESTful client.

> For python-icat the most obvious thing to do is to access both
> the SOAP and rest APIs and use whichever is most appropriate for
> your call.  You already access the IDS using a rest API.

Yes, python-icat uses the IDS RESTful interface.  But IDS is very simple and the main purpose is file up- and download.  This is indeed one use case where I see a point in using RESTful, because file transfer would be difficult to realize with SOAP in a clean way.  But did you ever take a look into the icat.ids module of python-icat?  How clumsy this code is, how much tedious boilerplate code you need to add for composing and decoding the messages, even for the most trivial task?  For SOAP, all this is done automatically by the underlying library.  The library can do this, because it understands the structure of these messages, which would be impossible with RESTful.

> The maxEntities property is needed by the IDS so that it does not
> produce calls to ICAT that will fail and the same is true for
> anybody else using ICAT.  I agree that lifetimeMinutes is of less
> value as you can log in and then see how much time you have left
> however I prefer to have direct access to the property.

I did not propose to change the `properties` call at the RESTful interface in any way.  My pull request does not touches this at all.  So IDS' use of maxEntities is not affected.

I only explained why my implementation of the `getAuthenticatorInfo` call at the SOAP interface is not completely conform to the corresponding RESTful `properties` call and that, if such a conformity is desired, my pull request would need to be amended.

> One of the benefits of the restful API is that the set of properties
> being returned can easily be modified without breaking existing code
> provided that this is done in a reasonable manner.

That is exactly the major disadvantage of RESTful.  If you change the structure of the messages that the server accepts or delievers, you will break the client.  With SOAP, such incompatibilities can be detected automatically, so that adding support for a new server version is seamless.  With RESTful, they will go unnoticed during server upgrade and will pop up sometime later during productive operation in the form of strange behavior of the client and bugs that are difficult to find and mostly impossible to understand.

PR #168, opened 2016-03-17
LOG table has been removed from ICAT 4.7 schema.

RKrahl:
LOG table has been removed from ICAT 4.7 schema, so I guess, it should also be removed by the upgrade script.


tomgriffinstfc:
I would say this needs to be migrated to dashboard rather than simply dropped.

fisherab:
I chose not to drop the table in case the user wants the information stored
there. I should add to the instructions that it may be deleted as it is no
longer required by ICAT.

Steve

RKrahl:
Ok.

PR #203, opened 2018-05-25
Create and destroy JMS Session and Producer on object startup and destroy. Fixes #202

### LINKED ISSUE #202: JMS Session not closed if an exception is thrown
### Description
There is a bug in the implementation of the [Transmitter](https://github.com/icatproject/icat.server/blob/master/src/main/java/org/icatproject/core/manager/Transmitter.java) and [NotificationTransmitter](https://github.com/icatproject/icat.server/blob/master/src/main/java/org/icatproject/core/manager/NotificationTransmitter.java) classes. If an exception is thrown whilst processing a JMS message, the Session object is not closed leading to all subsequent messages failing until the application is restarted. First reported by Jamie (@jsws).

@RKrahl: This also affects the IDS' [Transmitter](https://github.com/icatproject/ids.server/blob/master/src/main/java/org/icatproject/ids/Transmitter.java) class. 

### Symptoms

In the server.log, we see many warnings like this:

```
[2018-05-03T13:22:25.696+0100] [Payara 4.1] [WARNING] [] [javax.jms.Connection.mqjmsra] [tid: _ThreadID=9857 _ThreadName=http-thread-pool::http-listener-2(52)] [timeMillis: 1525350145696] [levelValue: 900] [[
  MQJMSRA_DC4001: createSession():An active Session object already exists for this connection, only one active Session object allowed per connection.]]
```

In the icat.log, we see many errors like this:

```
2018-05-03 00:04:40,511 ERROR [http-thread-pool::http-listener-2(2)] Transmitter - Failed to send jms message createMany 130.246.39.170
2018-05-03 00:04:40,516 ERROR [http-thread-pool::http-listener-2(2)] ICAT - Unexpected failure in Java 1.8.0_151
javax.jms.JMSException: MQJMSRA_DC4001: createSession():An active Session object already exists for this connection, only one active Session object allowed per connection.
        at com.sun.messaging.jms.ra.DirectConnection.checkSessionsAllowed(DirectConnection.java:1064) ~[imqjmsra.jar:na]
        at com.sun.messaging.jms.ra.DirectConnection.createSession(DirectConnection.java:398) ~[imqjmsra.jar:na]
        at org.icatproject.core.manager.NotificationTransmitter.processMessage(NotificationTransmitter.java:70) ~[NotificationTransmitter.class:na]
...
```

### Detail
The processMessage method on the Transmitter and NotificationTransmitter classes creates a Session object at the start and closes it at the end. If an exception is thrown in the method, it will never get to the close call and the Session object will remain. On the next call to the method, the Session creation will fail as there can only be one open Session at at time. The [Java documentation](https://docs.oracle.com/javaee/7/api/javax/jms/Connection.html#createSession-boolean-int-) states:
"Applications running in the Java EE web and EJB containers must not attempt to create more than one active (not closed) Session object per connection. If this method is called in a Java EE web or EJB container when an active Session object already exists for this connection then a JMSException may be thrown."

### Solution
One solution may be to wrap the contents of the method in a try-finally clause and put the close call in the 'finally' block. However, elsewhere (such as the [Apache ActiveMQ FAQ](http://activemq.apache.org/how-do-i-use-jms-efficiently.html)) it states that it is more efficient to create the objects at startup and reuse them.

Therefore I propose to refactor the method to move the creation of the objects (Session and MessageProducer) to the '@PostConstruct init' method and destruction to the '@PreDestroy exit'  method.

stuartpullinger:
Fixes #202

### DIFF
--- a/src/main/java/org/icatproject/core/manager/NotificationTransmitter.java
+++ b/src/main/java/org/icatproject/core/manager/NotificationTransmitter.java
@@ -33,6 +33,10 @@ public class NotificationTransmitter {
 
 	private TopicConnection topicConnection;
 
+        private Session jmsSession;
+
+        private MessageProducer jmsProducer;
+
 	@PostConstruct
 	private void init() {
 
@@ -42,6 +46,8 @@ public class NotificationTransmitter {
 					.lookup(propertyHandler.getJmsTopicConnectionFactory());
 			topicConnection = topicConnectionFactory.createTopicConnection();
 			topic = (Topic) ic.lookup("jms/ICAT/Topic");
+			jmsSession = topicConnection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+			jmsProducer = jmsSession.createProducer(topic);
 			logger.info("Transmitter created");
 		} catch (JMSException | NamingException e) {
 			logger.error(fatal, "Problem with JMS " + e);
@@ -53,6 +59,12 @@ public class NotificationTransmitter {
 	@PreDestroy()
 	private void exit() {
 		try {
+                        if (jmsProducer != null) {
+			        jmsProducer.close();
+                        }
+                        if (jmsSession != null) {
+			        jmsSession.close();
+                        }
 			if (topicConnection != null) {
 				topicConnection.close();
 			}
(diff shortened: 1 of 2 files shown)


RKrahl:
One additional remark: According to the Java EE 7 [Documentation on javax.jms.Session.close()](https://docs.oracle.com/javaee/7/api/javax/jms/Session.html#close--): “there is no need to close the producers and consumers of a closed session.” So, the explicit call of `jmsProducer.close()` added by this PR doesn't seem to be strictly necessary. It is certainly not harmful either.

PR #214, opened 2019-01-16
Use @Before and @After to setup and clear session in integration tests

louise-davies:
Some tests in TestRS were not properly initialising/clearing data and the order of tests being run caused tests to fail. Rather than initialising and clearing data in individual tests, refactor the code to use @Before and @After annotations to automatically perform these tasks before and after every test ran - ensuring better test seperation.


stuartpullinger:
Great work!

PR #221, opened 2019-03-13
Allow more types to be jsonised. Fixes #196. Fixes #220.

### LINKED ISSUE #196: EJBQL Serialization problem: Don't know how to jsonise class java.math.BigDecimal
Not sure if it is a problem related to ICAT but when I run queries like:
```
SELECT dataset.id, dataset.name, dataset.startDate, dataset.endDate, investigation.name, parameterType.name, parameter.stringValue, parameter.numericValue, ",
"(SELECT COUNT(p) from Investigation p) as investigationCount ",
"FROM DatasetParameter as parameter ",
"JOIN parameter.dataset dataset ",			   
"JOIN parameter.type parameterType ",
"JOIN dataset.investigation investigation ",
"where investigation.id = ?"
```

It produces an internal ICAT exception with the message:  Don't know how to jsonise class java.math.BigDecimal
```
2017-07-27 15:18:21,944 INFO [http-listener-2(24)] EntityBeanManager - Final search JPQL: SELECT  dataset.id , dataset.name , dataset.startDate , dataset.endDate , investigation.name , parameterType.name , parameter.stringValue , parameter.numericValue , ( SELECT COUNT ( p ) FROM Investigation p ) AS investigationCount FROM DatasetParameter AS parameter JOIN parameter.dataset dataset JOIN parameter.type parameterType JOIN dataset.investigation investigation WHERE (  investigation.id = '24332871' )
2017-07-27 15:18:22,129 ERROR [http-listener-2(24)] Transmitter - Failed to send jms message search 172.24.8.34
2017-07-27 15:18:22,132 ERROR [http-listener-2(24)] IcatExceptionMapper - Processing
org.icatproject.core.IcatException: Don't know how to jsonise class java.math.BigDecimal
	at org.icatproject.exposed.ICATRest.jsonise(ICATRest.java:744) ~[ICATRest.class:na]
	at org.icatproject.exposed.ICATRest.search(ICATRest.java:1186) ~[ICATRest.class:na]
```

### LINKED ISSUE #220: Cannot jsonise Study.StudyStatus
From @brianritchie1312:

> When I created a Study with a non-null status value, I got a 500 error: `Don't know how to jsonise field of type StudyStatus`. This comes from icat.server’s ICATRest.java, in a private method jsonise().  This uses field.getType().getSimpleName(); then deals with a small number of cases. Study’s status is an enum, StudyStatus, that is not recognised by this method. It works if the status value is null because that’s an initial catch-all case before the type test.

We should probably deal with #196 at the same time as this one. Perhaps adding some tests to make sure that all entities defined are able to be jsonised?

stuartpullinger:
This branch fixes a couple of problems converting some types into JSON and adds some tests.

### DIFF
--- a/src/test/java/org/icatproject/exposed/TestICATRest.java
+++ b/src/test/java/org/icatproject/exposed/TestICATRest.java
@@ -27,7 +27,6 @@ public class TestICATRest {
 
 	private static ICATRest icatRest = new ICATRest();
 	private static Method jsoniseArray;
-	private static Method jsoniseBean;
 
 	private final static DateFormat df8601 = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");
 
--- a/src/test/java/org/icatproject/exposed/TestICATRest.java
+++ b/src/test/java/org/icatproject/exposed/TestICATRest.java
@@ -1,0 +1,102 @@
+package org.icatproject.exposed;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+import org.junit.Test;
+import org.junit.BeforeClass;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.Throwable;
+import java.io.ByteArrayOutputStream;
+import java.util.Date;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+
+import javax.json.Json;
+import javax.json.stream.JsonGenerator;
+
+import org.icatproject.exposed.ICATRest;
+import org.icatproject.core.entity.ParameterValueType;
+import org.icatproject.core.entity.StudyStatus;
+import org.icatproject.core.IcatException;
+
+public class TestICATRest {
+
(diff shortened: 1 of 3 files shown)


src/test/java/org/icatproject/exposed/TestICATRest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.icatproject.exposed.ICATRest;
+import org.icatproject.core.entity.ParameterValueType;
+import org.icatproject.core.entity.StudyStatus;
+import org.icatproject.core.IcatException;
+
+public class TestICATRest {
+
+	private static ICATRest icatRest = new ICATRest();
+	private static Method jsoniseArray;
+	private static Method jsoniseBean;
```
louise-davies: Remove this as it is not used

src/test/java/org/icatproject/integration/TestRS.java
```diff
@@ -102,6 +107,109 @@ public void testDistinctBehaviour() throws Exception {
 
 	}
 
+	@org.junit.Rule
+        public ErrorCollector collector = new ErrorCollector();
```
louise-davies: Wrong indenting

PR #226, opened 2020-01-14
Python 3 compatibility

stuartpullinger:
This PR aims to allow the ICAT Server scripts to be run under Python 2 and Python 3.

Still need to:
- [x] Test on Windows + Python 2
- [x] Test on Windows + Python 3
- [x] Test on Linux + Python 2
- [x] Test on Linux + Python 3

### DIFF
--- a/src/main/scripts/testicat
+++ b/src/main/scripts/testicat
@@ -10,21 +10,21 @@ def check(r):
         if r.status_code / 100 != 2:
             json = r.json()
             if json.get("offset"):
-                print >> sys.stderr, json["code"], json["message"], json["offset"]
+                print(json["code"], json["message"], json["offset"], file=sys.stderr)
             else:
-                print >> sys.stderr, json["code"], json["message"]
+                print(json["code"], json["message"], file=sys.stderr)
             sys.exit(1)
 
 args = sys.argv
 if len(args) < 3 or len(args) % 2 != 1:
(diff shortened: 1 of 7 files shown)


l-mas:
testicat script is not compatable with python 2.7
```
File "src/main/scripts/testicat", line 13
    print(json["code"], json["message"], json["offset"], file=sys.stderr)
                                                             ^
SyntaxError: invalid syntax
```

Everythin else compiles on python 2 and 3

src/main/scripts/testicat
```diff
@@ -10,21 +10,21 @@ def check(r):
         if r.status_code / 100 != 2:
             json = r.json()
             if json.get("offset"):
-                print >> sys.stderr, json["code"], json["message"], json["offset"]
+                print(json["code"], json["message"], json["offset"], file=sys.stderr)
```
l-mas, line 13: This seems incompatable with python 2.7

RKrahl: I guess it will work if you add a
```python
from __future__ import print_function
```
statement as very first line of the script.

agbeltran: @l-mas @stuartpullinger has this been resolved?

RKrahl:
I just tested the `setup` script on Linux with both Python 2 and Python 3. Obviously this requires an updated version of `icat.utils` having icatproject/icat.utils#10 and icatproject/icat.utils#11 included to be used in the build. I only tested `setup install` and `setup uninstall` and found this to work.

louise-davies (approved):

Issue #200, opened 2018-01-02
ICAT schema: extensions to support data publication workflows


RKrahl:
According to its own claim, ICAT is supposed to link “all aspects of the research chain from proposal through to publication”. At HZB, we recently started to draft our data publication workflow. We want to manage data publications with ICAT. But we found the current schema would need some extensions.

We assume that a data publication in general may comprise one or more `Datasets` or `Datafiles` that may or may not belong to a single `Investigation`. The content of a publication would therefore best be represented by a `DataCollection`. There is currently no suitable place in the schema to store the publication metadata related to a `DataCollection`.

For the publication metadata, taking [DataCite 4.0](https://schema.datacite.org/meta/kernel-4.0/index.html) as orientation, the following attributes would be needed:
- Doi,
- Creator and contributor (`User`),
- Title,
- Publisher (`Facility`),
- Dates: date of creation and date of publication (Date of creation may be derived from the related datasets),
- Resource type (“Dataset”),
- Subject: keyword, classification code, or key phrase.
- Related identifiers. Relevant relation types may include: “IsCitedBy”, “Cites”, “IsSupplementTo”, “IsPartOf”, “IsCompiledBy”, “IsDerivedFrom”, “IsSourceOf”.
- Description, e.g. abstract.
- Funding information.

In order to create a suitable place to store this information, I suggest to add the following new classes to the schema:

## `DataPublication`

Represent a data publication.

Constraint: `facility`, `doi`

Relationships:

Card | Class               | Field
-----|---------------------|---------------------
1,1  | Facility            | facility
1,1  | DataCollection      | content
0,*  | DataPublicationUser | dataPublicationUsers
0,*  | RelatedIdentifier   | relatedIdentifiers
0,*  | FundingReference    | fundingReferences

Other fields:

Field           | Type
----------------|---------------------
doi             | String[255] NOT NULL
title           | String[255] NOT NULL
publicationDate | Date NOT NULL
subject         | String[1023]
description     | String[4000]

## `DataPublicationUser`

Author: creator of a or contributor to a data publication

Constraint: `publication`, `user`, `contributorType` 

Relationships:

Card | Class           | Field
-----|-----------------|------------
1,1  | DataPublication | publication
1,1  | User            | user

Other fields:

Field           | Type
----------------|---------------------
contributorType | String[255] NOT NULL
key             | String[255]
fullName        | String[255]
familyName      | String[255]
givenName       | String[255]
affiliation     | String[1023]

[...]

PR #232, opened 2020-05-05
ICAT schema extension for data publications

EmilJunker:
This adds the new classes proposed in #200 to the schema. The idea is to be able to store and manage information about data publications within ICAT.

Please note:

- I think I have done everything that (to the best of my knowledge) is necessary to add new classes to the schema. I have done some testing using python-icat and it seems to work fine. However, it is possible that I might be missing something since this is my first time working on the schema.
- #200 proposes `DataPublicationUsers` to have an attribute called 'key'. I have found that this can lead to all sorts of problems because KEY is also a keyword within JPQL queries. I have therefore changed the attribute's name to 'orderKey' for now.
- It is assumed that the changes in this PR will be released as part of icat.server version 4.11, hence that version number is used in the release notes and in the upgrade instructions.
- I have also created upgrade scripts for both MySQL and Oracle which create the new tables and constraints. I have successfully tested the MySQL script. The script for Oracle is pretty much the same, but I haven't tested that one, so I cannot say for sure if it actually works.


RKrahl:
Many thanks!

Just two minor remarks:

- I'm happy to rename the `DataPublicationUser.key` property to `orderKey`. I simply didn't consider that the name `key` is obviously prone to cause trouble.

- As discussed in the last meeting, the version to include the schema changes is expected to be 5.0. There are several schema changes pending and we would like to add them all at once in one version and so the changes will be significant enough to merit a major version bump.

RKrahl:
I tested this thoroughly.  I also created a modified version of my web service script providing the data publication landing pages. This version reads the publication metadata from ICAT using the schema extensions.

Everything works fine!

stuartpullinger:
Thanks for doing this. My first impression is that it looks good - I'd like to test it too, though. I have a couple of comments:

- Use of 'pid' and 'doi' as field names in ICAT is currently inconsistent. In the future, I think we had agreed that we should use 'pid/PID' everywhere as this is the more generic term and enables us to be consistent across all the entities. Do we need to discuss this in the group?
- The exact release that this will be merged into is not yet decided. If there is feature in this collection of schema extensions which it is difficult or problematic for a site to implement, I may split the features into 2 releases: easy-to-upgrade schema additions (such as this) in v4.11 and breaking schema changes in v.5.0.

RKrahl:
>     Use of 'pid' and 'doi' as field names in ICAT is currently inconsistent. In the future, I think we had agreed that we should use 'pid/PID' everywhere as this is the more generic term and enables us to be consistent across all the entities.

I'm happy to use `pid` instead of `doi`.

RKrahl:
Note: after incorporating the suggestions being made in the various discussions, the additions to the schema that this pull request now implemets look like:

## `DataPublication`

A curated data publication

Constraint: `facility`, `pid`

Relationships:

1,1  | Facility            | facility
1,1  | DataCollection      | content
0,*  | DataPublicationUser | users
0,*  | DataPublicationDate | dates
0,*  | RelatedIdentifier   | relatedIdentifiers
0,*  | FundingReference    | fundingReferences

Other fields:

Field           | Type                 | Description
----------------|----------------------|-----------------------------------------------
pid             | String[255] NOT NULL |
title           | String[255] NOT NULL | Title of the publication
publicationDate | Date                 | Date when the data was made publicly available
subject         | String[1023]         | List of keywords
description     | String[4000]         | Abstract

## `DataPublicationUser`

Author, e.g. creator of a or contributor to a data publication

Constraint: `publication`, `user`, `contributorType` 

Relationships:

Card | Class           | Field
-----|-----------------|------------
1,1  | DataPublication | publication
1,1  | User            | user

Other fields:

Field           | Type                 | Description
----------------|----------------------|---------------------------------------------------
contributorType | String[255] NOT NULL | See DataCite property contributorType or use "Creator"
orderKey        | String[255]          | Defines an order among the contributors
fullName        | String[255]          | May include title
familyName      | String[255]          | The family name of the user
givenName       | String[255]          | The given name of the user
affiliation     | String[1023]         | The home institute or other affiliation of the user
affiliationIdentifier | String[255] |

## `DataPublicationDate`

A date relevant for the publication

Constraint: `publication`, `dateType` 

Relationships:

Card | Class           | Field
-----|-----------------|------------
1,1  | DataPublication | publication

Other fields:

Field    | Type                 | Description
---------|----------------------|-----------------------------------------
dateType | String[255] NOT NULL | See DataCite property dateType
date     | String[255] NOT NULL | Use ISO 8601 format, may also be a range

(Suggested values for dateType include "Accepted", "Available", "Created", "Submitted".)

## `RelatedIdentifier`

Identifier of a related resource to a data publication

Constraint: `publication`, `identifier` 

Relationships:

Card | Class           | Field
-----|-----------------|------------
1,1  | DataPublication | publication

Other fields:

Field         | Type                 | Description
--------------|----------------------|-----------------------------------
identifier    | String[255] NOT NULL | The identifier of the related resource
relationType  | String[255] NOT NULL | See DataCite property relationType
fullReference | String[1023]

(Suggested values for relationType include "IsCitedBy", "IsCompiledBy", "IsReferencedBy", "IsSupplementTo", "References".)

## `FundingReference`

Information about financial support for a data publication

Constraint: `publication`, `funderName`, `awardNumber`

Relationships:

Card | Class           | Field
-----|-----------------|------------
1,1  | DataPublication | publication

Other fields:

Field            | Type
-----------------|---------------------
funderName       | String[255] NOT NULL
funderIdentifier | String[255]
awardNumber      | String[255] NOT NULL
awardTitle       | String[255]

RKrahl:
I submitted two new proposals that relate to data publications, #248 and #249.  We would need to decide on them before proceeding with this PR.  That is why I marked this one as draft.

PR #234, opened 2020-06-09
Add entity class Technique to the schema

### LINKED ISSUE #230: Add a entity class Technique
The [current draft for a data model](https://github.com/panosc-eu/search-api/blob/master/doc/data-model.md) for the PaNOSC WP3 search API envisages an entity class `Technique`. We currently do not have that in our schema. As discussed in the last F2F meeting, we might want to adopt this idea.

A `Technique` should have the attributes `name` and `pid` and have a many-to-many relation with `Dataset`. I therefore suggest to add the following new classes to the schema:

## `Technique`

Represents an experimental technique

Constraint: `facility`, `name`

Relationships:

Card | Class            | Field
-----|------------------|---------------------
1,1  | Facility         | facility
0,*  | DatasetTechnique | datasetTechniques

Other fields:

Field       | Type                 | Description
------------|----------------------|-------------------------------
name        | String[255] NOT NULL | A short name for the technique
pid         | String[255]          | A persistent identifier
description | String[255]          | An informal description

## `DatasetTechnique`

Represents a many-to-many relationship between a dataset and the experimental technique being used to create that Dataset

Constraint: `dataset`, `technique`

Relationships:

Card | Class     | Field
-----|-----------|----------
1,1  | Dataset   | dataset
1,1  | Technique | technique

EmilJunker:
This closes #230.


RKrahl:
Note: this pull request adopts the suggestions being made in the discussion and now implements the schema change in the version as described by @agbeltran [in the comment in #230](https://github.com/icatproject/icat.server/issues/230#issuecomment-652959419).

Issue #231, opened 2020-04-23
Make the relationship between Sample and Investigation many-to-many


RKrahl:
We have the `Sample` class in our schema to represent "a sample to be used in an investigation". It has many-to-one relationship with `Investigation`, e.g. each sample must be related to one and only one investigation.

While this is certainly suitable for the most situations, it maybe problematic in some cases: if one individual sample has been the subject of more then one investigation, there is currently no way to properly represent this in our schema. I'd like to remind that we occasionally also have rather prominent samples such as archeological artifacts or pieces of art and it is not so improbable that the same item might be investigated more then once by independent groups using different techniques trying to answer unrelated questions.

I therefore suggest to add the following class to the schema:

## `InvestigationSample`

Represents a many-to-many relationship between an investigation and a sample that has been used in that investigation

Constraint: `investigation`, `sample`

Relationships:

Card | Class         | Field
-----|---------------|----------
1,1  | Investigation | investigation
1,1  | Sample        | sample

As a result, the `samples` relationship in `Investigation` would be changed to have the type `InvestigationSample` rather then `Sample`. The `Sample` class would need to be modified as follows:

## `Sample`

A sample to be used in one or more investigations

Constraint: `facility`, `name`

Relationships:

Card | Class               | Field
-----|---------------------|----------
1,1  | Facility            | facility
0,*  | Dataset             | datasets
0,1  | SampleType          | type
0,*  | SampleParameter     | parameters
0,*  | InvestigationSample | investigationSamples

Other fields:

Field       | Type                 | Description
------------|----------------------|--------------------------------------------------
name        | String[255] NOT NULL |
pid         | String[255]          | A persistent identifier attributed to this sample

E.g. the many-to-one relation with `Investigation` would be replaced by a one-to-many relation with `InvestigationSample` and I suggest to add a relationship with `Facility` just for the sake of consistency with the current schema where most things are related directly or indirectly to the facility on top. The uniqueness constraint would need to be adapted as well. The attributes of `Sample` would remain unchanged.

kevinphippsstfc:
[...]

PR #240, opened 2020-07-06
Add many-to-many relationship InvestigationSample to the schema

EmilJunker:
This implements #231.

Here is a short summary of the changes:
- adds a new class `InvestigationSample` to the schema
  (representing a many-to-many relationship between `Investigation` and `Sample`)
- changes the unique constraint for the `Sample` table from `investigation, name` to just `pid`
  (adding a NOT NULL constraint for `pid` while also keeping the existing NOT NULL constraint for `name`)

**Note that this is still very much a work in progress.** A few points to note:

- The upgrade script initializes the value of the `Sample.pid` field with 'local:\<id\>' (reusing the `id` attribute) for all samples where the `pid` is currently `NULL`. However, in the current ICAT schema, the `pid` field has no unique constraint. This means that there could already be two samples with the same `pid` (at least theoretically). Or perhaps more likely: there could be several samples with the `pid` set to '' (an empty string). The upgrade script currently does not handle these cases (as they are very unlikely to occur in practice).

- The upgrade script also needs to move the existing one-to-many relationship between Investigation and Sample over to the new `InvestigationSample` table. Once this is done, the unique constraints on the `Sample` table are updated and the `Sample.investigation_id` column is dropped. I'm not sure if the current implementation of the upgrade script handles all of this correctly. In other words: thorough testing is needed.


EmilJunker:
I'm closing this PR in favor of #294.

Issue #248, opened 2021-03-10
Add new entity classes UserInfo and Affiliation


RKrahl:
Please note: for the moment, this is more an idea that would need thorough discussion, rather then a fully elaborated proposal.  In particular because it would be disruptive.

In #200, I proposed schema extensions to support data publication workflows.  This proposal was addressing a particular use case of making curated data publications.  It included a new entity `DataPublicationUser` that was needed because “affiliations and names of users may change over time. That's is why we need to keep track of affiliation and name at the time of the publication. Relying on the corresponding attributes of `User` only may not be enough.”

There was a longer discussion in that issue whether a `DataPublication` should be related to a `DataCollection` as proposed or to an `Investigation`. There was no conclusion, but I believe what @dfq16044 actually wanted to achieve was to get certain features out of this proposal to be applicable also to the different use case of simply minting DOIs for investigations instead.  One of these features she wanted to retain for that other use case was that ‘freezing’ of user attributes at the time of minting the DOI.

Now, if one contemplates more deeply about this, I believe the core of the problem is the double nature of the `User` class in our current schema.  At one hand, we need it to represent a person, in order to provide a means to authenticate that person, to manage access privileges and to record relations to other entities such as investigations.  On the other hand, we use it to store attributes like `givenName`, `familyName`, `email`, and `affiliation` of that person that are to be used in a particular context.  While the person always remains the same and we need a persistent record of that person, the attributes are subject to change and may need to be different according to the context.  Until now, we were only interested in the current state and the latest values for these attributes, so this did not become a problem.  The data publication was the first use case that needed to keep a record of the attributes at some particular moment in time.  The discussion revealed that this is not the only case where this is needed and I suspect more use cases will come in the future.  So it seems, we need to address this issue fundamentally.

The idea for a solution would be to separate the two purposes of the `User` class: keep `User` to represent the person and add another class `UserInfo` for the potentially volatile attributes of that person.  More then one `UserInfo` instances may exist for the same `User`, but each `UserInfo` only relates to one `User`.  At the same time, there is another issue with `affiliation`: an author of a publication may need to have more then one affiliation in that publication.  So the proposal suggests the following classes:

## `User`

A user of the facility

Constraint: `name`

Relationships:

Card | Class     | Field
-----|-----------|-----------
0,*  | UserInfo  | infos
0,*  | UserGroup | userGroups

Other fields:

Field | Type                 | Description
------|----------------------|----------------------------------------------------------------------------
[...]

PR #256, opened 2021-09-27
ICAT schema extensions for ICAT Server 5.0 release

### LINKED ISSUE #200: ICAT schema: extensions to support data publication workflows

### LINKED ISSUE #211: Additional column for investigation and dataset sizes

### LINKED ISSUE #228: Add a relationship between Dataset and Instrument
In PaNOSC WP3 we are currently discussing the [API](https://github.com/panosc-eu/search-api) for the local metadata catalogues to be searched by the federated catalogue. While this discussion is not yet concluded, the [current draft for a data model](https://github.com/panosc-eu/search-api/blob/master/doc/data-model.md) envisages a relation between Dataset and Instrument.

In ICAT, we have that relation only indirectly via Investigation. This may be a problem, because the datasets within a single investigation may have been created with different instruments and there is no easy way to establish the information which dataset has been created at which instrument out of our current schema. For me it would make perfect sense to relate a dataset directly to the instruments that created it.

At least for HZB, that relationship should be many-to-many, because we have in general more then one single instrument being involved in a single measurement (e.g. beam line and experimental station).

### LINKED ISSUE #230: Add a entity class Technique

### LINKED ISSUE #238: Add volume and fileCount to Investigation and Dataset
It has always been a sort of issue to get the calculation with the total volume of a number of files. We have solved it by storing these values as DataParameter and InvestigationParameters and they are recalculated any time a file, dataset or investigation is modified in ICAT

We can live without but it might be interesting to consider to add them as fields of the tables Investigation and Dataset.

Example:

Field | Type | Description
-- | -- | --
name | String [255] NOT NULL | A short name for the investigation
startDate | Date |  
doi | String [255] | The Digital Object Identifier associated with this investigation
summary | String [4000] | Summary or abstract
modId | String |  
releaseDate | Date | When the data will be made freely available
visitId | String [255] NOT NULL | Identifier for the visit to which this investigation is related
createId | String |  
createTime | Date |  
modTime | Date |  
endDate | Date |  
title | String [255] NOT NULL | Full title of the investigation
volume | Integer | Total volume of the investigation 
fileCount | Integer | Total number of files of the investigation

### LINKED ISSUE #249: Add FundingReference class
This is similar in nature as #248.  I put put it up for discussion, although I'm not certain myself, which variant I would prefer.

In #200, I proposed schema extensions to support data publication workflows. This proposal was addressing a particular use case of making curated data publications. It included a new entity `FundingReference`.  But this was formulated in a way that it could only be used in relation with a `DataPublication`.  Now there may be other use cases where funding information may be needed.  This could be relevant when simply minting DOIs for investigations rather then making curated data publications.  But it may also be useful in general to collect this information already in the user office, when submitting the proposal and store it with the investigation.

So this would amend #200 in storing the `FundingReference` independently of `DataPublication`.  It would consist of the following classes:

## `FundingReference`

Information about financial support

Constraint: `funderName`, `awardNumber`

Relationships:

Card | Class                  | Field
-----|------------------------|---------------
0,*  | InvestigationFunding   | investigations
0,*  | DataPublicationFunding | publications

Other fields:

Field            | Type
-----------------|---------------------
funderName       | String[255] NOT NULL
funderIdentifier | String[255]
awardNumber      | String[255] NOT NULL
awardTitle       | String[255]

## `InvestigationFunding`

Relation between an investigation and a funding reference

Constraint: `investigation`, `funding`

Relationships:

Card | Class            | Field
-----|------------------|---------------
1,1  | Investigation    | investigation
1,1  | FundingReference | funding

## `DataPublicationFunding`

Relation between an data publication and a funding reference

Constraint: `dataPublication`, `funding`

Relationships:

Card | Class            | Field
-----|------------------|---------------
1,1  | DataPublication  | dataPublication
1,1  | FundingReference | funding

EmilJunker:
This PR collects (i.e. obsoletes) several existing PRs and provides a clean, conflict free way to merge them into master.

Obsoletes PR #232, closes #200 closes #249, includes parts of #248.
Obsoletes PR #233, closes #211, closes #238.
Obsoletes PR #234, closes #230.
Obsoletes PR #235, closes #228.


EmilJunker:
I've done some testing and it turned out that renaming `investigationSize` and `datasetSize` to `size` as proposed in https://github.com/icatproject/icat.server/pull/233#issuecomment-896870524 caused problems with Oracle because `SIZE` is apparently a registered keyword in PL/SQL. That is why I have now renamed the `size` attribute to `fileSize` for both Investigations and Datasets. This works well and is also in line with the `fileCount` attribute name.

RKrahl:
> I've done some testing and it turned out that renaming `investigationSize` and `datasetSize` to `size` as proposed in [#233 (comment)](https://github.com/icatproject/icat.server/pull/233#issuecomment-896870524) caused problems with Oracle because `SIZE` is apparently a registered keyword in PL/SQL. That is why I have now renamed the `size` attribute to `fileSize` for both Investigations and Datasets. This works well and is also in line with the `fileCount` attribute name.

Note: as an alternative to renaming the attribute, one could also override the column name in the database using the `Column` annotation. E.g. something like:
```java
public class Investigation extends EntityBaseBean implements Serializable {

	@Comment("The cumulated sizes of the datasets within this investigation")
	@Column(name = "SIZE_")
	private Long size;

	@Comment("The total number of datafiles within this investigation")
	private Long fileCount;
}
```

I'd be happy with either solution. To be discussed in the collaboration meeting.

RKrahl:
I get an error from the `upgrade_mysql_5_0.sql` script trying to upgrade from `icat.server` 4.11.1:
```
$ mysql icat < upgrade_mysql_5_0.sql
ERROR 1025 (HY000) at line 117: Error on rename of './icat/#sql-76_54' to './icat/AFFILIATION' (errno: 150 "Foreign key constraint is incorrectly formed")
```

RKrahl:
> I get an error from the `upgrade_mysql_5_0.sql` script trying to upgrade from `icat.server` 4.11.1:

There is an apparently related issue when deploying `icat.server` during a fresh install. I get the following  suspicious entry in the `server.log`:
```
[2021-11-18T11:34:04.093+0100] [Payara 4.1] [WARNING] [] [javax.org.glassfish.persistence.org.glassfish.persistence.common] [tid: _ThreadID=32 _ThreadName=admin-thread-pool::admin-listener(1)] [timeMillis: 1637231644093] [levelValue: 900] [[
  PER01000: Got SQLException executing statement "ALTER TABLE AFFILIATION ADD CONSTRAINT FK_AFFILIATION_DATAPUBLICATIONUSER_ID FOREIGN KEY (DATAPUBLICATIONUSER_ID) REFERENCES DATAPUBLICATIONUSER (ID)": java.sql.SQLException: Can't create table `icat`.`AFFILIATION` (errno: 150 "Foreign key constraint is incorrectly formed")]]
```

Apart from that log entry, the installation seem to succeed. I believe that `icat.server` simply ignores SQL errors during install. But if one inspects the SQL dump after installation, it turns out that the foreign key constraint is indeed missing from the `AFFILIATION` table:
```sql
CREATE TABLE `AFFILIATION` (
  `ID` bigint(20) NOT NULL AUTO_INCREMENT,
  `CREATE_ID` varchar(255) NOT NULL,
  `CREATE_TIME` datetime NOT NULL,
  `MOD_ID` varchar(255) NOT NULL,
  `MOD_TIME` datetime NOT NULL,
  `NAME` varchar(1023) NOT NULL,
  `PID` varchar(255) DEFAULT NULL,
  `DATAPUBLICATIONUSER_ID` bigint(20) NOT NULL,
  PRIMARY KEY (`ID`),
  UNIQUE KEY `UNQ_AFFILIATION_0` (`DATAPUBLICATIONUSER_ID`,`NAME`) USING HASH
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

RKrahl:
> I get an error from the `upgrade_mysql_5_0.sql` script trying to upgrade from `icat.server` 4.11.1:

Ok, I found it.

I got more details about that foreign key error running `SHOW ENGINE INNODB STATUS` in the database:
```
2021-11-18 11:21:13 0x7f38402a8700 Error in foreign key constraint of table icat/AFFILIATION:
there is no index in the table which would contain
the columns as the first columns, or the data types in the
table do not match the ones in the referenced table
or one of the ON ... SET NULL columns is declared NOT NULL. Constraint:
,
  CONSTRAINT `FK_AFFILIATION_DATAPUBLICATIONUSER_ID` FOREIGN KEY (`DATAPUBLICATIONUSER_ID`) REFERENCES `DATAPUBLICATIONUSER` (`ID`)
```

Actually, there is an index in the table which contains `DATAPUBLICATIONUSER_ID` as the first column, but it is declared `USING HASH`, see above. Apparently such an index is unusable for a foreign key, which is kind of logic. It seems that the reason for the `USING HASH` declaration was that the `name` attribute having 1023 characters is too wide to be used in an index directly. In any case, if one reduces the name attribute to 511 characters (which I guess, will still be enough), everything works.

RKrahl:
This PR adds or modifies the following entity classes:

# New classes

## `Affiliation`

The home institute or other affiliation of a user in the context of a data publication

**Uniqueness constraint**: `user`, `name`

Relationships:

1,1  | DataPublicationUser | user

Other fields:

name            | String [511] NOT NULL | Name and optionally address of the affiliation
pid             | String [255]          | Identifier such as ROR or ISNI

## `DataPublication`

A curated data publication

**Uniqueness constraint**: `facility`, `pid`

Relationships:

Card | Class                  | Field
-----|------------------------|---------------------
1,1  | Facility               | facility
1,1  | DataCollection         | content
0,*  | DataPublicationDate    | dates
0,*  | DataPublicationFunding | fundingReferences
0,*  | DataPublicationUser    | users
0,*  | RelatedIdentifier      | relatedIdentifiers

Other fields:

pid             | String [255] NOT NULL | Persistent Identifier of the publication, such as a DOI
title           | String [255] NOT NULL | Title of the publication
publicationDate | Date                  | Date when the data was made publicly available
description     | String [4000]         | Abstract
subject         | String [1023]         | List of keywords

## `DataPublicationDate`

A date relevant for the publication

**Uniqueness constraint**: `publication`, `dateType`

Relationships:

1,1  | DataPublication     | publication

Other fields:

dateType        | String [255] NOT NULL | Type of the date, see DataCite property dateType for suggested values
date            | String [255] NOT NULL | Use ISO 8601 format, may also be a range

## `DataPublicationFunding`

Represents a many-to-many relationship between a data publication and a funding reference

**Uniqueness constraint**: `dataPublication`, `funding`

Relationships:

1,1  | FundingReference    | funding
1,1  | DataPublication     | dataPublication

## `DataPublicationUser`

Author, e.g. creator of a or contributor to a data publication

**Uniqueness constraint**: `publication`, `user`, `contributorType`

Relationships:

1,1  | DataPublication     | publication
1,1  | User                | user
0,*  | Affiliation         | affiliations

Other fields:

contributorType | String [255] NOT NULL | Role of that user in the publication, see DataCite property contributorType for suggested values or use "Creator"
orderKey        | String [255]          | Defines an order among the contributors
givenName       | String [255]          | The given name of the user
familyName      | String [255]          | The family name of the user
fullName        | String [255]          | May include title

## `DatasetInstrument`

Represents a many-to-many relationship between an instrument and a dataset with data collected at that instrument

**Uniqueness constraint**: `dataset`, `instrument`

Relationships:

1,1  | Dataset             | dataset
1,1  | Instrument          | instrument

## `DatasetTechnique`

Represents a many-to-many relationship between a dataset and the experimental technique being used to create that Dataset

**Uniqueness constraint**: `dataset`, `technique`

Relationships:

1,1  | Dataset             | dataset
1,1  | Technique           | technique

## `FundingReference`

Information about financial support

**Uniqueness constraint**: `funderName`, `awardNumber`

Relationships:

Card | Class                  | Field
-----|------------------------|---------------------
0,*  | DataPublicationFunding | publications
0,*  | InvestigationFunding   | investigations

Other fields:

Field            | Type                  | Description
-----------------|-----------------------|-----------------------------------------------
funderName       | String [255] NOT NULL | Name of the funding entity
funderIdentifier | String [255]          | Unique identifier of the funding entity, such as a Crossref Funder ID
awardNumber      | String [255] NOT NULL | Code assigned by the funder to identify the grant, suggest to use ":unas" if there is no such identifier
awardTitle       | String [255]          | Title or name of the grant

## `InvestigationFunding`

Represents a many-to-many relationship between an investigation and a funding reference

**Uniqueness constraint**: `investigation`, `funding`

Relationships:

1,1  | Investigation       | investigation
1,1  | FundingReference    | funding

## `RelatedIdentifier`

Identifier of a related resource to a data publication

**Uniqueness constraint**: `publication`, `identifier`

Relationships:

1,1  | DataPublication     | publication

Other fields:

identifier      | String [255] NOT NULL | The identifier of the related resource
relationType    | String [255] NOT NULL | Description of the relationship with the related resource, see DataCite property relationType for suggested values
fullReference   | String [1023]         | The full reference for the related resource as it should be displayed on the landing page

## `Technique`

Represents an experimental technique

**Uniqueness constraint**: `name`

Relationships:

0,*  | DatasetTechnique    | datasetTechniques

Other fields:

name            | String [255] NOT NULL | A short name for the technique
pid             | String [255]          | A persistent identifier attributed to the technique, ideally referring to a vocabulary term
description     | String [255]          | An informal description for the technique

# Modified classes

Not mentioned here: entity classes that are the target of a many-to-one relation from one of the new classes get the corresponding reverse one-to-many relation added.

## `Dataset`

New fields:

fileCount       | Long                  | The total number of datafiles within this dataset
fileSize        | Long                  | The cumulated sizes of the data files within this dataset

## `Investigation`

New fields:

fileCount       | Long                  | The total number of datafiles within this investigation
fileSize        | Long                  | The cumulated sizes of the datasets within this investigation

src/test/java/org/icatproject/integration/TestRS.java
```diff
@@ -500,7 +500,7 @@ private JsonArray searchDatafiles(Session session, String user, String text, Dat
 	@Test
 	public void testVersion() throws Exception {
 		ICAT icat = new ICAT(System.getProperty("serverUrl"));
-		assertTrue(icat.getVersion().startsWith("4."));
+		assertTrue(icat.getVersion().startsWith("5."));
```
kevinphippsstfc, line 503: This comment is not about this line but more about whether we should be adding some tests that create and use the new entities and relationships created? 

I'm new to looking at the icat.server code but I can see that the current level of testing done is not ideal and not all existing entities appear to be tested, so I was wondering what other people think about this?

kevinphippsstfc (changes requested):

src/main/java/org/icatproject/core/entity/FundingReference.java
```diff
@@ -0,0 +1,91 @@
...
+	private List<DataPublicationFunding> publications = new ArrayList<>();
+
+	@Comment("Name of the funding entity")
+	@Column(name = "FUNDERNAME", nullable = false)
+	private String funderName;
+
+	@Comment("Unique identifier of the funding entity, such as a Crossref Funder ID")
+	private String funderIdentifier;
+
+	@Comment("Code assigned by the funder to identify the grant, suggest to use \":unas\" if there is no such identifier")
```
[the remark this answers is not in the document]
RKrahl: It means "value unassigned". The idea is to use the value as is when creating DataCite metadata out of the information in ICAT.  The DataCite schema defines standard values for unknown information, see Appendix 3 in the [DataCite Schema documentation](https://doi.org/10.14454/3W3Z-SA82).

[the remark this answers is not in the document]
kevinphippsstfc: OK, thanks. Perhaps we should add the words "from the Datacite schema" after :unas to make it clearer what it refers to.

src/main/java/org/icatproject/core/entity/DataPublication.java
```diff
@@ -0,0 +1,157 @@
...
+	private List<DataPublicationUser> users = new ArrayList<>();
+
+	@OneToMany(cascade = CascadeType.ALL, mappedBy = "publication")
+	private List<DataPublicationDate> dates = new ArrayList<>();
+
+	@OneToMany(cascade = CascadeType.ALL, mappedBy = "publication")
+	private List<RelatedIdentifier> relatedIdentifiers = new ArrayList<>();
+
+	@OneToMany(cascade = CascadeType.ALL, mappedBy = "dataPublication")
+	private List<DataPublicationFunding> fundingReferences = new ArrayList<>();
```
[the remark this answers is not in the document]
RKrahl: I believe the abbreviated attribute names (`publication` to refer to a `DataPublication`) are justified here.

The new entity classes `DataPublicationDate`, `DataPublicationUser`, and `RelatedIdentifier` are centered around `DataPublication`.  Objects in these classes are entirely dependent on the publication they relate to, they are meaningless without it.  It is obvious from this context that an attribute named `publication` in these classes can only refer to the `DataPublication`.  Overly verbose attribute names do not improve readability of the code.  This goes in the same way as the `type` attribute in `Investigation` refers to the `InvestigationType` and not to a `SampleType` and the `parameters` in `Dataset` refer to `DatasetParameters`.

The situation is slightly different for `DataPublicationFunding`, because that classes establishes the relation between a `DataPublication` and a `FundingReference`.  A `FundingReference` object does have a meaningful existence independent from `DataPublication` in the new ICAT schema.

So I'd prefer leaving the `publication` attributes in `DataPublicationDate`, `DataPublicationUser`, and `RelatedIdentifier`.

src/main/java/org/icatproject/core/entity/DataPublicationUser.java
```diff
@@ -0,0 +1,117 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.FetchType;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+import javax.persistence.UniqueConstraint;
+
+@Comment("Author, e.g. creator of a or contributor to a data publication")
```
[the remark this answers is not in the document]
RKrahl: Done.

src/main/java/org/icatproject/core/entity/Dataset.java
```diff
@@ -36,6 +36,12 @@
 	@OneToMany(cascade = CascadeType.ALL, mappedBy = "dataset")
 	private List<Datafile> datafiles = new ArrayList<Datafile>();
 
+	@Comment("The cumulated sizes of the data files within this dataset")
```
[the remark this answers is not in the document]
RKrahl: Done.

src/main/java/org/icatproject/core/entity/Investigation.java
```diff
@@ -29,6 +29,12 @@
 	@OneToMany(cascade = CascadeType.ALL, mappedBy = "investigation")
 	private List<Dataset> datasets = new ArrayList<>();
 
+	@Comment("The cumulated sizes of the datasets within this investigation")
```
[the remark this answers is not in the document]
RKrahl: Done.

src/main/java/org/icatproject/core/entity/FundingReference.java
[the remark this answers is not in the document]
RKrahl: Done.

src/test/java/org/icatproject/integration/TestRS.java
RKrahl: That is a general problem in `ìcat.server`: we do not have a complete test suite covering all features and the full schema.  Adding that would be a new project.  It would require a systematic approach and the design of a comprehensive test strategy.  I don't think this can be fixed in a PR considering only a subset of the schema.

src/main/java/org/icatproject/core/entity/DataPublication.java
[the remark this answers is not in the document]
kevinphippsstfc: OK. If you're comfortable with it I don't feel strongly enough about it to argue the case for it to be changed.

src/test/java/org/icatproject/integration/TestRS.java
kevinphippsstfc: I agree with your comments. This is a wider problem and not something that we are going to resolve here, or even improve the situation much by just adding tests for these new additions.

kevinphippsstfc (approved):

RKrahl:
I mentioned in the meeting last Thursday that there might be an issue with new entity classes and the export and import calls in `icat.server`.  I checked this and was proven wrong: `icat.server` turns out to be better than I thought, it just works, even for the new classes.

There is however one issue with the `fileCount` and `fileSize` attributes in `Dataset` and `Investigation`. If the database triggers are in places, the values will be wrong after an export / import cycle.

kevinphippsstfc:
Thanks @RKrahl. That's good news about the export and import just working for the new classes.

It's a shame about the `fileCount` and `fileSize` fields. Can this be fixed or would it be too much work to be worth it?

My feeling is that if it can't be fixed easily then we have to live with that and probably just add a warning to any documentation that (hopefully!) exists on the export/import functionality. The workaround after doing an import would presumably be to either re-calculate all the values using the `INITIALIZE_DS/INV_SIZE_COUNT` procedures in the ICAT 5 migration scripts (assuming you have a fairly small ICAT), or to edit the procedure to only update the Investigations that you have modified if you have a large ICAT and it's too expensive to re-calculate all values in the whole ICAT.

RKrahl:
> It's a shame about the `fileCount` and `fileSize` fields. Can this be fixed or would it be too much work to be worth it?

I'd say, it can't easily be fixed. It is due to the design decisions we made for the `fileCount` / `fileSize` feature: we decided to implement the feature with triggers in the database backend and to make installing the triggers optional. That means, `icat.server` has no way to know whether the triggers are in place. The triggers do not to recalculate the values but do only incremental updates when a `Datafile` is created, deleted or updated. Recalculating would have been too expensive performance-wise. And we decided not to enforce correct values on create or update of `Database` and `Investigation` objects. So it's always possible to create a `Dataset` or an `Investigation` with "wrong" values.

Now, what happens is: the export writes `Investigations` and `Datasets` with their (presumably correct) attribute values to the export file. The import reads that file and creates `Investigations` and `Datasets` with the (correct) values they have set in the file. Then the import reads the `Datafiles` from the file and creates them. This causes the database triggers to update `fileCount` and `fileSize` in the corresponding `Dataset` and `Investigation` objects. *bummer*: as result, the values are doubled.

> My feeling is that if it can't be fixed easily then we have to live with that and probably just add a warning to any documentation that (hopefully!) exists on the export/import functionality.

Yes.

> The workaround after doing an import would presumably be to either re-calculate all the values

That would be my advise. As an alternative, one could temporarily disable the triggers during the import.

> […] if you have a large ICAT and it's too expensive to re-calculate all values in the whole ICAT.

I don't believe import or export are ever used on a large production ICAT.

src/main/java/org/icatproject/core/entity/DataPublicationDate.java
```diff
@@ -0,0 +1,58 @@
...
+	@JoinColumn(name = "DATAPUBLICATION_ID", nullable = false)
+	@ManyToOne(fetch = FetchType.LAZY)
+	private DataPublication publication;
+
+	@Comment("Type of the date, see DataCite property dateType for suggested values")
+	@Column(name = "DATETYPE", nullable = false)
+	private String dateType;
+
+	@Comment("Use ISO 8601 format, may also be a range")
+	@Column(nullable = false)
```
kevinphippsstfc, line 28: To make this work on Oracle where `DATE` is a reserved word, we are going to need to add something like `name = "DATE_"` to the definition of this column. I've tested this on MariaDB and Oracle and it works. The `upgrade_mysql/oracle_5_0.sql` scripts will need the column name updating as well.

src/main/scripts/upgrade_oracle_5_0.sql
```diff
@@ -0,0 +1,205 @@
...
+	CREATE_TIME TIMESTAMP NOT NULL,
+	MOD_ID VARCHAR2(255) NOT NULL,
+	MOD_TIME TIMESTAMP NOT NULL,
+	DATASET_ID NUMBER(19) NOT NULL,
+	TECHNIQUE_ID NUMBER(19) NOT NULL,
+	PRIMARY KEY (ID)
+);
+
+ALTER TABLE DATASETTECHNIQUE ADD CONSTRAINT FK_DATASETTECHNIQUE_DATASET_ID FOREIGN KEY (DATASET_ID) REFERENCES DATASET (ID);
+ALTER TABLE DATASETTECHNIQUE ADD CONSTRAINT FK_DATASETTECHNIQUE_TECHNIQUE_ID FOREIGN KEY (TECHNIQUE_ID) REFERENCES TECHNIQUE (ID);
```
kevinphippsstfc, line 31: I don't think we specify anywhere what versions of Oracle ICAT supports but if we want to remain compatible with versions prior to 12.2 where the length of an Oracle identifier was increased from 30 to 128 bytes, then we will need to reduce the length of some of these constraint names. From a quick bit of googling and reading of https://en.wikipedia.org/wiki/Oracle_Database and https://support.oracle.com/knowledge/Oracle%20Database%20Products/742060_1.html it looks like 11.2 and 12.1 are still being supported.

kevinphippsstfc (changes requested):

src/main/scripts/upgrade_oracle_5_0.sql
RKrahl: I don't mind changing that.  But this is only the upgrade script.  If you do a fresh install the database initialization code that JPA implicitly generates from the Java Source will use the same constraint names (I believe).

kevinphippsstfc: It's only the upgrade script that needs changing. For a fresh install, Eclipselink is presumably aware of this limitation in Oracle because it reduces the length of the constraint name accordingly. 

Out of interest, it seems to drop the "FK" from the beginning, drop the underscores and then start dropping vowels until the length is 30 characters. 

I could copy the autogenerated names that Eclipselink chooses into our upgrade script, if you like, or we could try to give them better names ourselves. I don't think it matters too much either way, but let me know if you have a preference.

kevinphippsstfc: Actually, just thinking about this a bit more, I think I prefer the option to use the Eclipselink generated names. That way, if we need to write upgrade scripts in the future that modify any of the constraints then at least we know what name they should have in all ICATs (freshly installed ones or upgraded ones).

RKrahl: I'd suggest that we should try to keep the result of the upgrade script as close as possible to a fresh install.  Since I do not test on Oracle myself, if you could change the upgrade script for Oracle to match the autogenerated names that would be great.

kevinphippsstfc: 👍

src/main/java/org/icatproject/core/entity/DataPublicationDate.java
kevinphippsstfc: @RKrahl If you are happy with the re-naming of the data column to DATE_ then I'm happy to add this to my pull request along with the change required for the constraint names.

src/main/scripts/upgrade_oracle_5_0.sql
kevinphippsstfc: I have now made these changes on the `icat-schema-extension` branch which we are looking to merge here.

kevinphippsstfc (approved):

RKrahl:
I now merged #260 into this branch and updated the comment describing the new classes above accordingly.

RKrahl:
I now merged the other pending PRs requesting schema changes into this one. From my point of view, it's ready. I did not test this with Oracle, though. @kevinphippsstfc, could you pleas check again.

kevinphippsstfc:
Thanks @RKrahl. I will test this branch on Oracle.

kevinphippsstfc:
I've done a `mvn install` with the the test setup using an Oracle database and the tests all pass apart from:
`TestRS.testSearch:691->search:1034 expected:<4> but was:<5>`
which is the test that PR #275 aims to fix.
Did you manage to take a look at that yet?

kevinphippsstfc:
I've done a `mvn install` with the the test setup using an Oracle database and the tests all pass apart from:
`TestRS.testSearch:691->search:1034 expected:<4> but was:<5>`
which is the test that PR #275 aims to fix.
Did you manage to take a look at that yet?

kevinphippsstfc (approved):

PR #274, opened 2022-02-04
Remove authn.simple from run.properties

### LINKED ISSUE #272: Remove need for authn.simple from integration tests
Even though the tests only use authn.db, the run.properties file set up by prepare_test.py specifies that authn.simple should also be installed.

For now, the need for authn.simple should be removed.

In the longer term, I would prefer to use authn.simple rather than authn.db as it is simpler to set up, but this will need to be coordinated with changes to icat-ansible to get the CI tests passing, so this is currently not as simple as it might appear.

kevinphippsstfc:
Closes #272

### DIFF
--- a/src/test/scripts/prepare_test.py
+++ b/src/test/scripts/prepare_test.py
@@ -27,14 +27,13 @@ if not os.path.exists("src/test/install/run.properties"):
     with open("src/test/install/run.properties", "w") as f:
         contents = [
             "lifetimeMinutes = 120",
-            "rootUserNames = db/root simple/root",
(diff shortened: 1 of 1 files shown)


RKrahl:
It should be noted that the `prepare_test.py` script hard codes many things in the properties files that may or may not make sense depending on the test environment. At least in my environment, I need to manually edit the files in `src/test/install` after they have been created by the script anyway. So, there is still room for improvement in the way how we setup the tests. Nevertheless, the changes made in this PR are useful.

RKrahl (approved):

Issue #277, opened 2022-02-21
Speed up filterReadAccess for free text searches


patrick-austin:
## Current behaviour
Currently, when performing a free text search using Lucene, we get results in batches of 1000, then iterate over these one by one:
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/manager/EntityBeanManager.java#L784-L808

Performing authorisation first checks for root access, public tables, and failing these, gets and evaluates rules relevant to that single entity's ID:
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/entity/Rule.java#L39
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/entity/Rule.java#L165-L166

## Desired behaviour
There is an existing method which instead takes authorises based on several entities at once:
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/manager/GateKeeper.java#L167-L246
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/entity/Rule.java#L40
https://github.com/icatproject/icat.server/blob/f86b255fc7ab43cdcb55876b57966c22ff39b08a/src/main/java/org/icatproject/core/entity/Rule.java#L167-L168

We should use this for the free text search, as we already get results in batches, and reducing the number of queries to the DB for rules should speed up the free text search component without affecting the results returned.

PR #278, opened 2022-02-21
Add getReadableIds and use in filterReadAccess #277

patrick-austin:
# Changes
Added `getReadableIds` function.  The logic here is the same as in `getReadable`, however we accept and return lists of IDs rather than entities. When dealing with the free text search results, we only have IDs. Before we were finding the bean for each ID in turn in order to run authorisation on it, then proceeding to return the ID. But as we already have the `klass` of the entitity and the ID from Lucene we can simply operate on these instead.

Instead of hardcoding the number of results to get from Lucene at 1000, instead use the `maxIdsInQuery` setting. This is the limit we would send in one batch to the DB in `getReadable`/`getReadableIds` anyway, and the default value of 500 and max for Oracle of 1000 is the same order of magnitude as what we were currently using.

# Impact
I ran a number of searches against my local development components, using the [lils.yml ](https://raw.githubusercontent.com/icatproject-contrib/icat.ansible/master/files/lils.yml) as my dummy data. This contains:
| Investigations | Datasets | Datafiles |
|---------------:|----------:|---------:|
|100 | 300 | 29967|

I searched for `*` (returning all entities before authorising) as both `root` and the non-root `icatuser`, representing best and worst case scenarios (all data visible, no data visible and needing to check DB for rules) for the old and new setup with `maxIdsInQuery` set to 500 and 1000. All times are in ms, and is measured between the first `EntityBeanManager - Got X results from Lucene` and `EntityBeanManager - Returning Y results` logs made by icat.server.

## root
| Setup | Investigations | Datasets | Datafiles |
|:-------|---------------:|----------:|---------:|
| Old     | 91 | 205 | 212 | 
| New (500)| 15 | 11 | 12 |
| New (1000) | 13 | 12 | 11 |

Old method is comparable for DS and DF since we only loop until we reach 300 accepted results, then return.
New only submits one batch of entities to check, and returns in consistently ~12ms.

## non-root

| Setup | Investigations | Datasets | Datafiles |
|:-------|---------------:|----------:|---------:|
| Old     | 215 | 337 |29261 | 
| New (500)| 21 | 14 | 1502 |
| New (1000) | 15 | 14 | 690 |

Old method takes around 1 to 2 ms per entity checked as we need to check every ID individually before being sure none are authorised.
New method takes roughly the same amount of time for Investigations and DS as these both require a single batch of IDs (less than the limit in settings. Datafiles takes around half the time when the limit is doubled to 1000, as we only need to send half as many batches (averaging 24ms per batch).

Also searched for a specific single result using `id:...` in both old and new setup as root:
| Setup | Investigations | Datasets | Datafiles |
|:-------|---------------:|----------:|---------:|
| Old     | 17 | 14 | 16 | 
| New  | 11 | 15 | 14 |

Times are comparable with old and new approaches. For the new method, the time taken to authorise 1 result and hundreds (as in the `*` example) was also comparable.

It's worth noting that in these tests I didn't actually have any rules in the DB to evaluate.

### DIFF
--- a/src/main/java/org/icatproject/core/manager/EntityBeanManager.java
+++ b/src/main/java/org/icatproject/core/manager/EntityBeanManager.java
@@ -148,7 +148,7 @@ public class EntityBeanManager {
 	private boolean luceneActive;
 
 	private int maxEntities;
-	private int maxIdsInQuery;
+	private int luceneSearchBlockSize;
 
 	private long exportCacheSize;
 	private Set<String> rootUserNames;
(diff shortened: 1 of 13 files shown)


src/main/java/org/icatproject/core/manager/EntityBeanManager.java
```diff
@@ -781,27 +782,37 @@ private void exportTable(String beanName, Set<Long> ids, OutputStream output,
 		}
 	}
 
-	private void filterReadAccess(List<ScoredEntityBaseBean> results, List<ScoredEntityBaseBean> allResults,
+	private void filterReadAccess(List<ScoredEntityBaseBean> acceptedResults, List<ScoredEntityBaseBean> newResults,
```
kevinphippsstfc, line 785: Could you add add a method comment here please, in particular the relationship between acceptedResults and newResults. I realise there was no comment before but I like the fact that you have given the arguments better names and I think just a bit more explanation would help future developers.

src/main/java/org/icatproject/core/manager/GateKeeper.java
```diff
@@ -164,35 +164,102 @@ private void exit() {
...
-
-		if (beans.size() == 0) {
-			return beans;
-		}
-
-		EntityBaseBean object = beans.get(0);
-
-		Class<? extends EntityBaseBean> objectClass = object.getClass();
-		String simpleName = objectClass.getSimpleName();
+	/**
```
kevinphippsstfc, line 167: Add a method description as well, perhaps using some of the description for the return type.

src/main/java/org/icatproject/core/manager/GateKeeper.java
```diff
@@ -164,35 +164,102 @@ private void exit() {
...
+		for (EntityBaseBean bean : beans) {
+			if (readableIds.contains(bean.getId())) {
+				results.add(bean);
 			}
 		}
+		return results;
+	}
+
+	/**
+	 * @param userId     The user making the READ request
```
kevinphippsstfc, line 232: Add a method description as well, perhaps using some of the description for the return type.

kevinphippsstfc (changes requested):
Overall, this looks great, thanks! I've just requested a few additions to the comments to hopefully help future developers.

I'm wondering how best to test this given that I don't think the part of the code using "restrictions" will get tested much (if at all?) by the unit and integration tests. We could do with testing it on an ICAT containing some of the usual InvestigationUser and InstrumentScientist rules/restrictions.

patrick-austin:
> Overall, this looks great, thanks! I've just requested a few additions to the comments to hopefully help future developers.
> 
> I'm wondering how best to test this given that I don't think the part of the code using "restrictions" will get tested much (if at all?) by the unit and integration tests. We could do with testing it on an ICAT containing some of the usual InvestigationUser and InstrumentScientist rules/restrictions.

Have added/expanded the documentation of the new/modified functions, if anything still needs clarifying just let me know.

Issue #254, opened 2021-08-18
Add a Role entity and relations from InstrumentScientist and InvestigationUser


stuartpullinger:
Add a `Role` table with label and pid and use it for InstrumentScientist and InvestigationUser. This will also require two more tables InstrumentScientistRole and InvestigationUserRole to allow one InstrumentScientist or InvestigationUser to have many roles. This implies removing the role field in the InvestigationUser and InvestigationGroup.

This solution was agreed at a meeting of the ICAT collaboration on 18th August 2021. This solution would replace the solution in #237.

RKrahl:
A comment on this, because I raised the question in today's meeting, without having carefully read the proposal before: if you add a many-to-many relation between `InvestigationUser` and `Role` (e.g. adding `InvestigationUserRole`), my concern about removing the `role` attribute from the uniqueness constraint in `InvestigationUser` is resolved. I was concerned because we have users with multiple roles in one investigation. But this will still be possible.

The only issue that I can see now is that the database upgrade script for this will be far from trivial: at the moment, we have multiple `InvestigationUser` entries between a user and an investigation having different `role` values. The upgrade script will need to merge them into a single one, e.g. retain one `InvestigationUser`, populate the `Role` table with multiple related entries accordingly and finally delete all the others. Putting that into a piece of correct SQL sounds tricky.

kevinphippsstfc:
Now that we are looking at the ICAT schema changes more closely and planning our upgrades to it, we have realised that for Diamond in particular this change could potentially cause us problems in the near future due to a temporary lack of effort available to work on the User Office to ICAT propagation component which would need modifying.

What Diamond do need however in the near future is the schema changes for `DataPublication`. So I can see two possible options here:

1. We delay this Role change until a future release. (Can somebody confirm who needs this change and when they would need it by please?)
2. We keep the `role` attribute in `InvestigationUser` for now (as well as adding the new relationship to `InvestigationUserRole`) but mark it as deprecated such that it will be removed in a future release. This would allow facilities to be more flexible about when they migrate from using the `role` attribute to using the new relationship.

Does anybody have any preferences or comments on this?

@antolinos I'm mentioning you because you requested this change in #237 but it doesn't look like you are connected to this issue yet.

antolinos:
@kevinphippsstfc No problem from my side to delay

RKrahl:
[...]

PR #291, opened 2022-05-24
Drop SQL upgrade scripts

RKrahl:
As discussed in the meeting  on 19th May, we might want to drop the SQL upgrade scripts and rely on the initialization in `icat.server` provided by EclipseLink to upgrade the database instead.

This PR changes the `eclipselink.ddl-generation` setting from `create-tables` to `create-or-extend-tables` and drops the SQL upgrade scripts for version 5.0. The EclipseLink setting has the effect that the initialization code will not only add missing tables, but also add missing columns to existing tables.

Note that this works for simple additive schema changes as we had in recent versions and we envision for 5.0. More complex schema changes such as those discussed in #231 or #254 can not be covered automatically by the EclipseLink initialization and will still require dedicated upgrade scripts.

I tested this with MariaDB for the upgrade from 4.11.1 to 5.0 (the current icat-schema-extension branch). I even tested an upgrade from `icat.server` 4.7.0 to 5.0 in one single step without running any upgrade scripts. This works as well with one noteworthy exception: the EclipseLink initialization did not create the index on `Datafile.location` [added in version 4.8](https://github.com/icatproject/icat.server/commit/186082e87952890a5af00361f9a319a9a0fb5e9f). But our SQL upgrade scripts didn't care to do that either.

@kevinphippsstfc, could you please test this with Oracle?


kevinphippsstfc (approved):
I tested this on Oracle and compared the database schemas (tables, indexes, constraints) and they were the same for a fresh 5.0.0-SNAPSHOT install as for a 4.11.1 schema that had been upgraded by the installation of the 5.0.0-SNAPSHOT. 

I also ran the integration tests against the Oracle database.

kevinphippsstfc (changes requested):
I realised that the installation HTML file also needs updating to reflect the fact that there are no longer upgrade files but that the schema upgrade for this version is automatic upon installation.

RKrahl:
> I realised that the installation HTML file also needs updating to reflect the fact that there are no longer upgrade files but that the schema upgrade for this version is automatic upon installation.

I wonder how we should deal with the schema upgrade instructions for very old (older than 4.7.0) versions. These would still require a stepwise upgrade from version to version with manual interventions in each step as described in the [install instructions in 4.7.0](https://repo.icatproject.org/site/icat/server/4.7.0/installation.html#Schema_upgrade) (and still in the current release). Should we keep these instructions?

I'd assume that in the meanwhile nobody is using that old versions in production anyway. So I'd be inclined to take the opportunity to drop these old instructions and simply state that direct upgrade is supported for ICAT 4.7.0 and newer. For older versions, one should first upgrade to 4.7.0 following the instructions provided with that version and then upgrade to 5.0.0 in a second step. Upgrade from 4.7.0 or newer is covered by the automatic initialization provided by EclipseLink.

kevinphippsstfc (approved):
I agree it is time to remove the upgrade instructions for the really old versions and I think the way you have done it makes sense.

PR #297, opened 2022-07-26
Add attributes to data publication related classes

### LINKED ISSUE #295: Add email to DataPublicationUser
While testing the `icat.server` snapshot release, I noticed that we missed something: we would also need an `email` attribute in `DataPublicationUser`. The reason is the same as for `givenName`, `familyName` and `affiliations`: we may want to include this information in the publication metadata or on the publication landing page, but they are subject to change, so we can't rely on the attributes in `User`, we need to keep a record of this information at the moment of the publication.

For `email`, this may be even more critical, this can be illustrated best with an [example of a real data publication](https://doi.org/10.5442/ND000003).  On the landing page, the email addresses of two of the authors are exposed to allow for getting into contact with them.  It was a conscious decision of the authors in this case to designate the 2nd and the 4th author to serve as a contact for this publication, but not the others. So we'd need a way to store this decision.

### LINKED ISSUE #296: Add relatedItemType and title to RelatedIdentifier
The latest [DataCite Schema version 4.4](https://schema.datacite.org/meta/kernel-4.4/) added the `RelatedItem` property.  It is in some way an extended version of `RelatedIdentifier` with additional attributes.  Now, I became aware that it actually may be beneficial to use `RelatedItem` instead of or additional to `RelatedIdentifier` in the publication metadata.  But in order to do that, we would need two more attributes in `RelatedIdentifier` in the ICAT schema:

Field           | Type        | Description
----------------|-------------|----------------------------------
relatedItemType | String[255] | The type of the related item
title           | String[255] | Title or name of the related item

(Obviously, the names are subject to discussion, we could for instance also put `relatedItemTitle`. For now, I just retained to corresponding property names from DataCite.)

RKrahl:
Add some more optional attributes to data publication related classes.

In detail:

* Add the following attribute to `DataPublicationUser`:
  Field | Type        | Description
  ------|-------------|---------------------------------------------------------------------------------
  email | String[255] | The email address for the user that should be exposed in the publication, if any

* Add the following attributes to `RelatedIdentifier`:
  Field           | Type        | Description
  ----------------|-------------|--------------------------------------
  relatedItemType | String[255] | The type of the related resource
  title           | String[255] | Title or name of the related resource

Close #295 and close #296.

### DIFF
--- a/src/main/java/org/icatproject/core/entity/DataPublication.java
+++ b/src/main/java/org/icatproject/core/entity/DataPublication.java
@@ -42,7 +42,7 @@ public class DataPublication extends EntityBaseBean implements Serializable {
 	private List<DataPublicationDate> dates = new ArrayList<>();
 
 	@OneToMany(cascade = CascadeType.ALL, mappedBy = "publication")
-	private List<RelatedIdentifier> relatedIdentifiers = new ArrayList<>();
+	private List<RelatedItem> relatedItems = new ArrayList<>();
 
 	@OneToMany(cascade = CascadeType.ALL, mappedBy = "dataPublication")
 	private List<DataPublicationFunding> fundingReferences = new ArrayList<>();
@@ -91,8 +91,8 @@ public class DataPublication extends EntityBaseBean implements Serializable {
 		return dates;
 	}
 
-	public List<RelatedIdentifier> getRelatedIdentifiers() {
-		return relatedIdentifiers;
+	public List<RelatedItem> getRelatedItems() {
+		return relatedItems;
 	}
 
 	public List<DataPublicationFunding> getFundingReferences() {
@@ -139,8 +139,8 @@ public class DataPublication extends EntityBaseBean implements Serializable {
 		this.dates = dates;
 	}
 
-	public void setRelatedIdentifiers(List<RelatedIdentifier> relatedIdentifiers) {
-		this.relatedIdentifiers = relatedIdentifiers;
+	public void setRelatedItems(List<RelatedItem> relatedItems) {
+		this.relatedItems = relatedItems;
 	}
 
 	public void setFundingReferences(List<DataPublicationFunding> fundingReferences) {
--- a/src/main/java/org/icatproject/core/entity/DataPublicationUser.java
+++ b/src/main/java/org/icatproject/core/entity/DataPublicationUser.java
@@ -47,6 +47,9 @@ public class DataPublicationUser extends EntityBaseBean implements Serializable 
 	@Comment("The family name of the user")
 	private String familyName;
 
+	@Comment("The email address for the user that should be exposed in the publication, if any")
+	private String email;
+
 	/* Needed for JPA */
 	public DataPublicationUser() {
 	}
@@ -83,6 +86,10 @@ public class DataPublicationUser extends EntityBaseBean implements Serializable 
 		return familyName;
 	}
 
+	public String getEmail() {
+		return email;
+	}
+
 	public void setPublication(DataPublication publication) {
 		this.publication = publication;
 	}
@@ -113,5 +120,9 @@ public class DataPublicationUser extends EntityBaseBean implements Serializable 
 
 	public void setFamilyName(String familyName) {
 		this.familyName = familyName;
+	}
+
+	public void setEmail(String email) {
+		this.email = email;
 	}
 }
--- a/src/main/java/org/icatproject/core/entity/RelatedIdentifier.java
+++ b/src/main/java/org/icatproject/core/entity/RelatedIdentifier.java
@@ -32,10 +32,10 @@ public class RelatedIdentifier extends EntityBaseBean implements Serializable {
 	@Column(length = 1023)
 	private String fullReference;
 
-	@Comment("The type of the related item")
+	@Comment("The type of the related resource")
 	private String relatedItemType;
 
-	@Comment("Title or name of the related item")
+	@Comment("Title or name of the related resource")
 	private String title;
 
 	/* Needed for JPA */
--- a/src/main/java/org/icatproject/core/entity/RelatedIdentifier.java
+++ b/src/main/java/org/icatproject/core/entity/RelatedIdentifier.java
@@ -32,6 +32,12 @@ public class RelatedIdentifier extends EntityBaseBean implements Serializable {
 	@Column(length = 1023)
 	private String fullReference;
 
+	@Comment("The type of the related item")
+	private String relatedItemType;
+
+	@Comment("Title or name of the related item")
+	private String title;
+
 	/* Needed for JPA */
 	public RelatedIdentifier() {
 	}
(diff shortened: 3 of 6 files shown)


kevinphippsstfc (commented):
The change to add the email is fine but I don't think adding the two additional fields for RelatedItem (in DataCite) to RelatedIdentifier is the right thing to do.

Reading the text in the DataCIte 4.4 spec about the addition of RelatedItem it says: "a new property, relatedItem, is introduced which contains information about a resource related to the one being registered, e.g., a journal article, conference paper, or a book series. It can be used to describe a text citation where relatedIdentifier cannot be used because the related resource does not have an identifier"

So a RelatedItem does not have an identifier, but we have `identifier` as a mandatory field in RelatedIdentifier. Also, in RelatedItem `title` is a mandatory field but in RelatedIdentifier title is optional.

Which makes me feel that we really should have an additional entity for RelatedItem with the mandatory and optional fields set correctly rather than trying to use RelatedIdentifier for both purposes. 

I may be missing something here. I'm not very familiar with the DataCite schema and I don't find it easy to interpret, but let me know what you think.

kevinphippsstfc:
I notice that the CI is failing. I didn't have time to investigate that so could you take a look.

RKrahl:
The `RelatedItem` in DataCite 4.4 is an extended and generalized version of `RelatedIdentifier`. It is extended in that it has additional subproperties, such as `relatedItemType`, `Creator`, `Title`, `PublicationYear`, … The purpose of these properties is mostly to allow generating a citation line referencing the related resource. They could in principle be fetched from the metadata of the related identifier, at least if that identifier is a DataCite DOI. It is generalized in the sense that the `relatedIdentifier` is optional, so it is also possible to reference items that do not have an identifier.

To give a concrete example: if a dataset is published with a DataCite DOI and the data has been collected from an instrument, that instrument could be referenced in the DataCite metadata using `RelatedIdentifier` as:
```xml
<relatedIdentifier relatedIdentifierType="DOI" relationType="IsCompiledBy">10.5442/NI000001</relatedIdentifier>
```
or using `RelatedItem` as:
```xml
<relatedItem relatedItemType="Other" relationType="IsCompiledBy">
  <relatedItemIdentifier relatedItemIdentifierType="DOI">10.5442/NI000001</relatedItemIdentifier>
  <titles>
    <title>E2 - Flat-Cone Diffractometer</title>
  </titles>
</relatedItem>
```
or using both redundantly.  The `RelatedItem` provides additional information, the type of the related resource[^1] and the name of the instrument, included as `title`. The same information could be fetched from the referenced [instrument DOI metadata](https://api.datacite.org/application/vnd.datacite.datacite+xml/10.5442/ni000001). Note that, as one can see in this example, `RelatedItem` has a subproperty `relatedItemIdentifier`, so it's not only for the case that the case of a related resource having no identifier.

Adding the additional information directly in the DataCite metadata may be needed, for instance if the metadata should be harvested by B2FIND. B2FIND needs to map the incoming metadata onto the [EUDAT Core Metadata Schema](http://eudat7-ingest.dkrz.de/guidelines/mapping.html#b2fmdschema). EUDAT Core has a text property for `Instrument`, so it would be possible to map the `title` from the `RelatedItem` onto that.[^2] But (as I learned last week) the mapper from B2FIND does not support resolving external identifiers and fetching metadata from the related DOIs. So if the instrument would only be linked using `RelatedIdentifier`, it would not be possible to have the `Instrument` property set in B2FIND.

In order to be able to use `RelatedItem`, we would need at least `relatedItemType` and `title`, as proposed in this PR, because these subproperties are mandatory.

[^1]: That type is `Other` here, because there is no `Instrument` in the controlled list of terms for resource type in the current DataCite version. But hopefully it will be added come in the future.
[^2]: The mapper does support individual per repository configuration. So it would be possible to add a rule: if a dataset relates a resource with `IsCompiledBy`, then that resource should be taken as an instrument.

RKrahl:
… continued:

> So a RelatedItem does not have an identifier, but we have `identifier` as a mandatory field in RelatedIdentifier. Also, in RelatedItem `title` is a mandatory field but in RelatedIdentifier title is optional.

As explained above, in DataCite, `RelatedItem` may have an identifier, but as opposed to `RelatedIdentifier`, it is not mandatory. And I hope, I made clear that there are good reasons to use `RelatedItem` also for resources that have an identifier.

I'm not so sure if there is any use case for linking to resources not having an identifier in the context of a data publication from a PaN facility. So I don't believe, the fact that `identifier` is mandatory in `RelatedIdentifier` in the ICAT schema would be an issue in practice.

> Which makes me feel that we really should have an additional entity for RelatedItem with the mandatory and optional fields set correctly rather than trying to use RelatedIdentifier for both purposes.

I believe, having separated classes in the ICAT schema for both cases would make things overly complicated. In practice, one might want to add both properties for the same related resource in the DataCite metadata: `RelatedItem`, because it provides the additional subproperties needed for instance for B2FIND and `RelatedIdentifier` for backward compatibility, because `RelatedItem` is relatively new and might not be understood by all consumers of the metadata.

In the end, it will be a site specific script that generates the DataCite metadata out of ICAT, either the script that generates the data publication landing pages or the XSLT file that generates the metadata in `icat.oaipmh`. The schema as proposed in this PR allows for all options one might want to implement. It is for instance relatively easy to code into the XSLT something like:

* always generate a `RelatedIdentifier` entry out of `identifier` and `relationType`
* additionally generate a `RelatedItem` entry if `relatedItemType` and `title` are not empty.

A site that doesn't care about `RelatedItem` may just ignore `relatedItemType` and `title`.

RKrahl:
> I notice that the CI is failing. I didn't have time to investigate that so could you take a look.

The CI is always failing for any things I submit. I guess that is a permission issue. I also don't get to see any diagnostic messages, so I can't tell what is going wrong.

RKrahl:
As discussed with @kevinphippsstfc today, we decided to rename the entity class `RelatedIdentifier` to `RelatedItem` with this PR in order to make it clearer what this is supposed to be. For the same reason, I expanded some of the comment strings to provided additional hints what should be put into the attributes.

So the new class now looks like:

--- 

## `RelatedItem`

A reference to an external resource or item that is related to a data publication, such as a scientific article that is based on the data or the instrument that has been used to collect the data

**Uniqueness constraint**: `publication`, `identifier`

Relationships:

1,1  | DataPublication     | publication

Other fields:

identifier      | String [255] NOT NULL | The identifier of the related resource
relationType    | String [255] NOT NULL | Description of the relationship with the related resource, see DataCite property relationType for suggested values
fullReference   | String [1023]         | The full reference for the related resource as it should be displayed on the landing page
relatedItemType | String [255]          | The type of the related resource, see DataCite property resourceTypeGeneral for suggested values
title           | String [255]          | Title or name of the related resource

 --- 

Obviously, the corresponding one-to-many relation in `DataPublication` has also been renamed accordingly.

kevinphippsstfc (approved):
I updated the release notes as well to refer to RelatedItem instead of RelatedIdentifier

Issue #304, opened 2023-03-09
TestRS exportMetaDataDump brittle wrt id length


patrick-austin:
## Description
The integration tests calling https://github.com/icatproject/icat.server/blob/edd2c1ebf915836176b346d4fa3835a1c75530a8/src/test/java/org/icatproject/integration/TestRS.java#L1501
import the test data (`icat.port`), dump it as either `root` or a normal user (as `dump1.txt`). We then import from `dump1.txt`, and dump again into `dump2.txt`. Finally, the length (not the contents) of the two files is compared.

Rule, DataCollection and Job all appear in the dump with their id. This increments as the tests run. The test will fail if during it, the number of digits in the ID crosses to a new order of magnitude. For example, 38 Rules are created each time `setAuthz` is called, so if dump1 contains:
```
...
"962","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublication"
"963","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublicationType"
"964","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublicationDate"
```
dump2 will contain:
```
...
"1000","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublication"
"1001","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublicationType"
"1002","db/root",2023-03-08T18:03:35.000Z,"db/root",2023-03-08T18:03:35.000Z,"CRUD","notroot","DataPublicationDate"
```
And the test will fail because dump2 is three characters longer than dump1.

When/if this will happen is highly dependent on how many tests are in TestRS, and how many times `setAuthz` is called.

## Context
I do not get these failures locally running the tests, as the DB used by my development instance is on id ~55720. When running the CI, it will use a fresh DB and go from 0 to ~1000 during a single execution of the tests. This failure may also be dependent on the DB backend used.

## Solutions
[...]

PR #306, opened 2023-04-14
Fix icatadmin to run with Python 3

### LINKED ISSUE #305: The icatadmin script fails to run with Python 3
Running the `icatadmin` script to populate the Lucene index with Python 3 fails:
```
$ python3 /opt/payara5/bin/icatadmin http://icat:8080 simple username root password secret -- populate
a bytes-like object is required, not 'str'
```
Trying the same with Python 2 works:
```
$ python2 /opt/payara5/bin/icatadmin http://icat:8080 simple username root password secret -- populate
Datafile
DatafileParameter
Dataset
DatasetParameter
Investigation
InvestigationParameter
InvestigationUser
Sample
```

Note that the 5.0.0 release of `icat.server` ships `icatadmin` with `#!/usr/bin/env python` in the first line, which will cause it to be executed with Python 2 by default on most systems. So the issue may go unnoticed in production by now.  But as a result of #300, this line has been changed to `#!/usr/bin/env python3`, so the error becomes apparent in development versions and future releases.

RKrahl:
Debugging `icatadmin` for #305 reveals, the issue occurs in line 117:
```python
    if parameters and method == "POST":
        conn.send(parameters)
```
`parameters` is the result of a `urllib.parse.urlencode()` call, which returns a string.  `conn` is either a `http.client.HTTPConnection` or a `http.client.HTTPSConnection`, `conn.send()` seem to require a bytes-like object rather than a string (the official Python docu is not very clear in that point, but experiemts confirm this). So the obvious fix is to convert the string to bytes.

Fix #305.

Note that I only implemented a minimal fix. There are other lines of code that feed strings into `conn.send()`, as in:
```python
    elif body:
        blocksize = 8192
        datablock = body.read(blocksize)
        crc32 = 0
        while datablock:
            conn.send(hex(len(datablock))[2:] + "\r\n")
            conn.send(datablock + "\r\n")
            crc32 = zlib.crc32(datablock, crc32)
            datablock = body.read(blocksize)
        conn.send("0\r\n\r\n")
```
I didn't bother to fix those, because:

1. `icatadmin` is awkward and should probably rather be rewritten from scratch.
2. those lines cited above are actually never executed, because the keyword arguments `headers` and `body` to `_process()` are never used and thus `body` will always be `None`.

### DIFF
--- a/src/main/scripts/icatadmin
+++ b/src/main/scripts/icatadmin
@@ -114,7 +114,7 @@ def _process(relativeUrl, parameters, method, headers=None, body=None):
     conn.endheaders()
      
     if parameters and method == "POST":
-        conn.send(parameters)
+        conn.send(parameters.encode("ascii"))
     elif body:
         blocksize = 8192
         datablock = body.read(blocksize)
--- a/src/main/scripts/icatadmin
+++ b/src/main/scripts/icatadmin
@@ -133,7 +133,7 @@ def _process(relativeUrl, parameters, method, headers=None, body=None):
             responseContent = response.read()
             om = json.loads(responseContent)
         except Exception:
-            fatal("InternalException " + responseContent)
+            fatal("InternalException " + responseContent.decode())
         code = om["code"]
         message = om["message"]
         fatal(code + " " + message)
@@ -152,7 +152,7 @@ def getPopulating(args):
     try:
         sessionId = getService()
         parameters = {"sessionId": sessionId}
-        print(_process("lucene/db", parameters, "GET").read())
+        print(_process("lucene/db", parameters, "GET").read().decode())
     except Exception as e:
         fatal(e)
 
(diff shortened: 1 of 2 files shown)


patrick-austin (commented):
I checked and adding `encode` fixes the issue for me.

I did a bit more experimenting and found that `populating` would return `b'[]'` with Python3 instead of `[]` with Python2. Adding a corresponding `decode` after `read` fixed this for me. When we catch the exception we also need to `decode` to avoid a TypeError from attempting to concatenate a str with bytes (line 136).

I also added these changes to rules.py (although I couldn't get past an error about certificates to properly test their impact). A lot of the code is similar so (if anyone uses rules.py) it would benefit from the same fix(es).

I agree with the sentiment about avoiding a larger change due to a wider potential need for rewrites (at the very least linting/formatting/refactoring across icatadmin and rules.py).

The test failure on you original commit seemed to be due to #304 as far as I could tell.

If you're OK with my additional changes I'd be happy to merge this for the 5.0.1.

RKrahl:
Dear @patrick-austin, you are completely right, many thanks for your fixes!   I'm happy with your changes.

patrick-austin (approved):

PR #311, opened 2023-07-26
Update MySQL driver in setup.properties.example

RKrahl:
The MySQL section in the `setup.properties.example` file is outdated. The class `com.mysql.jdbc.jdbc2.optional.MysqlDataSource` does not exist any more in recent versions of `mysql-connector-java`. The driver needs to be updated to `com.mysql.cj.jdbc.MysqlDataSource`.

I don't know exactly at which version of `mysql-connector-java` this change has been made. But I believe, it's a while ago and at least any version that supports Java 11 requires the new class. So it should be safe to update that in the example file unconditionally.

### DIFF
--- a/src/main/config/setup.properties.example
+++ b/src/main/config/setup.properties.example
@@ -22,8 +22,7 @@ port           = 4848
 !db.logging     = FINE
 
 # MySQL
-db.driver      = com.mysql.jdbc.jdbc2.optional.MysqlDataSource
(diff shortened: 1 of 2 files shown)


patrick-austin:
The Python script that generates the `setup.properties` for the tests should also use the correct driver. The icat.manual tutorials should also be updated, I've been making some changes to those as I've been following them locally but will create an issue for it as well.

ajkyffin (approved):

PR #323, opened 2024-03-14
Fully-static EntityInfoHandler

ajkyffin:
In order to remove synchronized blocks, the PrivateEntityInfo instances are created during class static initialization and stored in immutable objects, making the class fully thread-safe.

Since everything is available in static variables, there is no need to instantiate the class, so all the methods are made static.

### DIFF
--- a/src/main/java/org/icatproject/core/DocGenerator.java
+++ b/src/main/java/org/icatproject/core/DocGenerator.java
@@ -45,9 +45,7 @@ public class DocGenerator {
 
 		for (String cname : cnames) {
 			out.print("<hr/><h2 id=\"" + cname + "\">" + cname + "</h2>");
-			Class<?> klass = Class.forName("org.icatproject.core.entity." + cname);
-			@SuppressWarnings("unchecked")
-			Class<? extends EntityBaseBean> eklass = (Class<? extends EntityBaseBean>) klass;
+			Class<? extends EntityBaseBean> eklass = EntityInfoHandler.getClass(cname);
 			String classComment = eiHandler.getClassComment(eklass);
 			if (classComment == null) {
 				System.out.println(cname + " has no comment");
--- a/src/main/java/org/icatproject/core/DocGenerator.java
+++ b/src/main/java/org/icatproject/core/DocGenerator.java
@@ -17,8 +17,6 @@ import org.slf4j.LoggerFactory;
 
 public class DocGenerator {
 
-	private static EntityInfoHandler eiHandler = EntityInfoHandler.getInstance();
-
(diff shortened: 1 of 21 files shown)


src/main/java/org/icatproject/core/DocGenerator.java
```diff
@@ -45,20 +43,18 @@ public static void main(String[] args) throws Exception {
...
 				System.out.println(cname + " has no comment");
 			} else {
 				out.println("<p>" + classComment + "</p>");
 			}
 
-			Set<Field> fields = eiHandler.getGetters(eklass).keySet();
-			Map<Field, String> fieldComments = eiHandler.getFieldComments(eklass);
-			Set<Field> notnullables = new HashSet<Field>(eiHandler.getNotNullableFields(eklass));
-			Map<Field, Integer> stringFields = eiHandler.getStringFields(eklass);
+			Set<Field> fields = new HashSet<Field>(EntityInfoHandler.getGetters(eklass).keySet());
```
patrick-austin, line 54: If `EntityInfoHandler.getGetters(eklass).keySet()` is already a `Set<Field>`, why do we create a new `HashSet`? Could this be:
```java
			Set<Field> fields = EntityInfoHandler.getGetters(eklass).keySet();
```



I was also thinking whether we could disambiguate and use `getFields`, i.e.:
```java
			Set<Field> fields = new HashSet<Field>(EntityInfoHandler.getFields(eklass));
```
But since the getters are build:
https://github.com/icatproject/icat.server/blob/c5ecf9d05d071678d221ae278a80a65769b3758d/src/main/java/org/icatproject/core/manager/EntityInfoHandler.java#L365-L382
before we remove the common fields
https://github.com/icatproject/icat.server/blob/c5ecf9d05d071678d221ae278a80a65769b3758d/src/main/java/org/icatproject/core/manager/EntityInfoHandler.java#L550-L556
these sets would be different - so probably why it used getters in the first place?

patrick-austin (commented):
Only found one thing, should work fine as is but wanted to check whether we can remove the (potentially redundant) `HashSet`, or not.

src/main/java/org/icatproject/core/DocGenerator.java
ajkyffin: The reason for the HashSet is that the Map returned by `EntityInfoHandler.getGetters()` is now unmodifiable, so a modifiable copy of the keySet is now needed (the `id` Field is removed from the Set a few lines down).

As for using `EntityInfoHandler.getFields()`:  I don't want to change any logic outside of EntityInfoHandler in this PR, especially if the results are likely to be different.

patrick-austin (approved):
Thanks for the explanation.

PR #346, opened 2025-03-12
Fix text preparation script: do not overwrite existing run.properties

RKrahl:
Depending on the environment when the test suite is run, it may be needed to tweak some parameters in the properties files.  For instance, the hard coded value for `search.directory` in the `run.properties` file may not always be appropriate in some environments. That is why the `prepare_test.py` takes care not to overwrite an existing `setup.properties` in `src/test/install`. This allows to override the settings there by putting a custom `setup.properties` file in that directory before running the tests. But the current scripts fails to take the same care for the `run.properties` file. This is fixed in the present PR.

### DIFF
--- a/src/test/scripts/prepare_test.py
+++ b/src/test/scripts/prepare_test.py
@@ -28,33 +28,34 @@ for f in glob.glob("src/test/install/*.war"):
 shutil.copy("src/main/config/run.properties.example",
             "src/test/install/run.properties.example")
 
-with open("src/test/install/run.properties", "w") as f:
-    contents = [
-        "lifetimeMinutes = 120",
-        "rootUserNames = db/root",
-        "maxEntities = 10000",
-        "maxIdsInQuery = 500",
-        "importCacheSize = 50",
-        "exportCacheSize = 50",
-        "authn.list = db",
-        "authn.db.url = %s" % icat_url,
-        "authn.simple.url = %s" % icat_url,
-        "notification.list = Dataset Datafile",
-        "notification.Dataset = CU",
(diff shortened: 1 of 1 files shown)


RKrahl:
Note on the failing CI test: I got the same error when a ran the tests locally first. It went away when I enabled the `units` setting in `icat.lucene`. I can't see what that setting is in the CI build.

patrick-austin (approved):
Thanks for identifying the improvement to `prepare_test.py` - I didn't notice there was different logic for `setup.properties`. I often manually add the "simple" authenticator to that script so I can use it for local development, but having it not overwrite a custom properties file that contains changes like that feels like a much better thing to do.

> Note on the failing CI test: I got the same error when a ran the tests locally first. It went away when I enabled the units setting in icat.lucene, so I guess the test suite needs that setting. I can't see what that setting is in the CI build.

The icat-ansible settings that are used to setup the CI did not include the latest settings for icat.lucene, including units, and were pulling the SNAPSHOT rather than the released version. I've made those changes: https://github.com/icatproject-contrib/icat-ansible/pull/114 and hopefully this will fix that test. We already pull that branch specifically in the icat.server CI, so we don't need to wait until that PR is merged, the next CI run should use those settings and we should be able to check if it has worked.

I think I missed this test when releasing v6.1.0 as the CI kept having intermittent failures with downloading all the components, e.g. https://github.com/icatproject/icat.server/actions/runs/13680392438/job/38250983297#step:13:299, so I was relying on the results of my local tests - which were using the correct icat.lucene settings.

PR #350, opened 2025-04-09
Fix CI Build workflow

RKrahl:
The CI Build workflow fails in step "Checkout icat-ansible" because it cannot checkout the `icat-6.1` branch of icat-ansible: this branch has been merged into master in icatproject-contrib/icat-ansible#114.


RKrahl:
> Given those failures aren't caused by this change I'm happy if you want to merge this first, or alternatively merge #351 onto this branch first, then merge this.

I don't have any particular preference on the order of the merges.  I now merged the `disable-db-triggers-ci` branch here, just to try out if everything works as expected.

PR #354, opened 2025-04-11
Improve queue file handling when indexing

ajkyffin:
- Limits the number of documents sent to icat.lucene in one go, in order to avoid timeouts
 - Commits the Lucene index after each batch of documents, in case there is a rollback in Lucene
 - Rotates the queue files if they get too big
 - Reads and writes from separate files to prevent writes timing out due to long-held locks

### DIFF
--- a/src/main/java/org/icatproject/core/manager/search/SearchManager.java
+++ b/src/main/java/org/icatproject/core/manager/search/SearchManager.java
@@ -70,6 +70,7 @@ public class SearchManager {
 	public static final int DEFAULT_INDEX_BATCH_SIZE = 500;
 	public static final int DEFAULT_INDEX_BATCHES_PER_TIMER = 10;
 	public static final int DEFAULT_BACKLOG_LINES_PER_TIMER = 10;
+	public static final long DEFAULT_QUEUE_FILE_MAX_SIZE = 10_000_000L;
 
 	public class EnqueuedSearchRequestHandler extends TimerTask {
 
@@ -542,6 +543,7 @@ public class SearchManager {
 	private int indexBatchSize;
 	private int indexBatchesPerTimer;
 	private int backlogLinesPerTimer;
+	private long queueFileMaxSize;
 
 	/**
 	 * Gets (and if necessary, builds) the fields which should be returned as part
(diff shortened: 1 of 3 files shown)


src/main/java/org/icatproject/core/manager/search/SearchManager.java
```diff
@@ -63,47 +67,104 @@
...
+						try (BufferedWriter writer = Files.newBufferedWriter(dotnewPath, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
+							String line;
+							while ((line = reader.readLine()) != null) {
+								writer.write(line);
+								writer.newLine();
+							}
+						} catch (IOException e) {
+							logger.error("Error writing to {}", dotnewPath, e);
+							return;
 						}
```
patrick-austin, line 131: Since this is inside the `while (numBatches < indexBatchesPerTimer)` loop, does this mean that we're potentially doing this write up to 10 tens per timer task execution? In practice there might not be much impact, but in principle it feels like were doing more IO than is needed - would there be a significant disadvantage/reason why we can't do the write/move after all the batches? I suppose it would be a bit less resilient since we have the `queue.N` file on disk with the full number of lines until the loop completes where in reality we already have either successfully committed these or written to the backLog (so we'd be double counting them in effect). If the impact of the extra writes is small then I suppose that's reason enough to do it this way?

src/main/java/org/icatproject/core/manager/search/SearchManager.java
```diff
@@ -145,25 +206,68 @@ private class PendingSearchRequestHandler extends TimerTask {
...
+						try (BufferedWriter writer = Files.newBufferedWriter(dotnewPath, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
+							String line2;
+							while ((line2 = reader.readLine()) != null) {
+								writer.write(line2);
+								writer.newLine();
+							}
+						} catch (IOException e) {
+							logger.error("Error writing to {}", dotnewPath, e);
+							return;
 						}
```
patrick-austin, line 246: Same as the point above, is the impact of doing this in the while loop small compared to the benefit of keeping the file on disk up to date/whatever other reasons we have for doing it this way?

patrick-austin (commented):
Can't see any problems with the code, I did want to check whether there's any significant factors for putting the write of unprocessed lines to `.new` inside the batch loop vs outside.

In terms of icat.lucene, did you still want to remove the commit timer as part of these changes? I couldn't see a PR for that yet. Also is there more pre-prod testing to do for this or has that been done?

ajkyffin:
I've changed it to write the .new files after processing all the lines/batches. A change to icat.lucene to stop the commit timer can come separately.

patrick-austin (approved):

PR #359, opened 2025-04-29
Add startDate and endDate fields to Instrument entity #349

### LINKED ISSUE #349: Add `startDate` & `endDate` fields to `Instrument`
Add optional `startDate` & `endDate` fields to the Instrument entity. This is to support creating DOIs for instruments & being able to add "Dates relevant to the instrument" like in the PIDINST schema (with the expectation that `startDate` represents the commission date and `endDate` represents the decommision date). Suggesting to use the names `startDate` & `endDate` to match other ICAT entities such as `Investigation` but open to other suggestions.

HZBPhilippA:
Add `startDate` & `endDate` fields to the `Instrument` entity to track instrument operational periods. These fields allow facilities to document when instruments began service and when they were decommissioned or are scheduled to end operation.

Close #349

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
 
 	<groupId>org.icatproject</groupId>
 	<artifactId>icat.server</artifactId>
-	<version>6.1.1-SNAPSHOT</version>
+	<version>6.2.0-SNAPSHOT</version>
 	<packaging>war</packaging>
 	<name>ICAT Server</name>
 	<description>A metadata catalogue to support Large Facility experimental data, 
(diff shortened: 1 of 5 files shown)


ajkyffin:
I don't think `Date` is the correct type to use here. It represents a specific time in UTC, whereas here we likely want just a calendar date. I think this would be better implemented as `String` so that you could have a more appropriate level of accuracy, e.g. `2025-05-01`, `2025-05-01T09:00:00` or `2025-05-01T09:00:00+02:00`. This would allow better mapping to other metadata schemas. What do you think @RKrahl?

ajkyffin:
If it's up to the client to interpret then that's fine, and consistency with other entities is big bonus.

PR #362, opened 2025-06-11
Remove EntityManager/UserTransaction/EJBs from some method signatures and standardise EntityManager variable name

ajkyffin:
Avoid passing instances of `EntityManager` and `UserTransaction` between EJB methods. These are interfaces to the application container's transaction management system and don't in themselves represent a database transaction. Passing them between EJB methods is functionally the same as injecting them into both EJBs, but could lead to misunderstanding how they work. `EntityManager` still needs to be passed to some entity methods because it cannot be injected into an entity.

Remove EJBs from method signatures where they are not used.

Renames all instances of `EntityManager` to `entityManager` instead of  `manager` for greater clarity.

### DIFF
--- a/src/main/java/org/icatproject/core/entity/DataCollectionParameter.java
+++ b/src/main/java/org/icatproject/core/entity/DataCollectionParameter.java
@@ -32,8 +32,8 @@ public class DataCollectionParameter extends Parameter implements Serializable {
 	}
 
 	@Override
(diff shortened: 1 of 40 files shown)


patrick-austin (commented):
Saw one place in `PublicStep.fixup()` where I thought we might not need the `EntityManager`?
Also one reference to the old name `manager` in a javadoc - most seem like they were updated apart from this one.

PR #363, opened 2025-06-13
Upgrade to JUnit 5

ajkyffin:
Upgrade from JUnit 4 to 5. Also upgrades the maven surefire and failsafe plugins for compatibility.

The main difference is that the optional message has moves from being the first argument to the last argument in assert methods.

This is required for quarkus because there is only a quarkus-junit5 extension.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,9 +90,16 @@
 		</dependency>
 
 		<dependency>
-			<groupId>junit</groupId>
-			<artifactId>junit</artifactId>
-			<version>4.13.2</version>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter</artifactId>
+			<version>5.13.1</version>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -90,9 +90,16 @@
...
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter</artifactId>
+			<version>5.13.1</version>
+			<scope>test</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-migrationsupport</artifactId>
+			<version>5.13.1</version>
```
patrick-austin, line 102: Is this needed/should it be left in? Or was it needed to do the migration but no longer required. I couldn't see any direct references to it in imports.

patrick-austin (commented):
Changes to test files look OK to me, just wanted to check if we need `junit-jupiter-migrationsupport` in the pom still or not?

PR #365, opened 2025-06-24
Increase location column size #364

### LINKED ISSUE #364: Increase size of location column
We are facing a problem with the length of the location field. We store the physical location, and sometimes (sporadically) it exceeds the 255-character limit.

It affects the Dataset and Datafile tables:
https://repo.icatproject.org/site/icat/server/6.1.0/schema.html#Dataset
https://repo.icatproject.org/site/icat/server/6.1.0/schema.html#Datafile

We propose considering extending the length to 512 (or other size 3000?)  characters by adding the decorator so both source code and DB scripts are generated accordingly. Example:
```
| @Comment("The logical location of the file - which may also be the physical location")
| @Column(length = 511)
| private String location;
`

HZBPhilippA:
Increased the maximum length of the location field in Dataset and Datafile entities from 255 to 511 characters.

Close #364


src/main/scripts/upgrade_mysql_6_1.sql
ajkyffin, line 1: Should this be `upgrade_mysql_6_2.sql`?

RKrahl: Renamed it now.

ajkyffin (approved):

PR #374, opened 2025-07-25
Add pid to sample type #326

### LINKED ISSUE #326: Add a pid property to SampleType and change the uniqueness constraint
The following proposal for a minor schema change is a result of a discussion with Kirsty Syder, Steve Collins and myself.  Credits should go to Steve for bringing it up.

The `molecularFormula` property in `SampleType` is not nullable, because it is part of the uniqueness constraint. This is unfortunate for some samples, maybe even for most, `molecularFormula` is meaningless. Of course we can work around this by setting something like `N/A` or `:unap` to denote "not applicable", but that is probably only the second best solution.

Ideally, the sample type should use a controlled vocabulary, such as an ontology. In this case, we'd want to link any `SampleType` object with the corresponding term in that vocabulary. Hence, we'd need to add a `pid` property to do that. Consequently, we should use that `pid` property also in the uniqueness constraint, which would allow to change `molecularFormula` to be nullable.

As an interim solution in the absence of any controlled vocabulary for sample types, we could populate the `pid` attribute with some arbitrary local id that only needs to be unique and is otherwise an opaque string.

HZBPhilippA:
A persistent identifier attributed to this sample type, ideally referring to a vocabulary term.

Close #326


RKrahl:
FIXME: `SampleType.description` is still missing.

PR #375, opened 2025-07-29
Isolate the Session table in a separate persistence context

ajkyffin:
This moves the `Session` table into its own persistence context in order to separate it from the metadata tables in the persistence layer. The session management methods are moved from `BeanManager` to `SessionManager`.

### DIFF
--- a/src/main/resources/META-INF/persistence.xml
+++ b/src/main/resources/META-INF/persistence.xml
@@ -9,6 +9,61 @@
 		<provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
 		<jta-data-source>jdbc/icat</jta-data-source>
 
+		<class>org.icatproject.core.entity.Affiliation</class>
+		<class>org.icatproject.core.entity.Application</class>
+		<class>org.icatproject.core.entity.DataCollectionDatafile</class>
+		<class>org.icatproject.core.entity.DataCollectionDataset</class>
+		<class>org.icatproject.core.entity.DataCollectionInvestigation</class>
+		<class>org.icatproject.core.entity.DataCollection</class>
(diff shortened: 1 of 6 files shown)


patrick-austin (approved):
Changes themselves look OK to me. Am I right in thinking we've been running this patched in (pre) prod for the facilities so it's been (manually) tested? If so happy for this to be merged.

RKrahl (approved):
I have to admit that I did not check all the changes in the sources in detail, I just skimmed over it and it looks reasonable.

But I build and tested the branch locally. I also ran the python-icat test suite against it, which as a side effect, is a rather comprehensive test for icat.server. I did not stumbled over any issue in this version.

I also tested that the change has the intended effect and can confirm that.

src/main/resources/META-INF/persistence.xml
```diff
@@ -9,6 +9,61 @@
...
+		<class>org.icatproject.core.entity.Sample</class>
+		<class>org.icatproject.core.entity.SampleParameter</class>
+		<class>org.icatproject.core.entity.SampleType</class>
+		<class>org.icatproject.core.entity.Shift</class>
+		<class>org.icatproject.core.entity.StudyInvestigation</class>
+		<class>org.icatproject.core.entity.Study</class>
+		<class>org.icatproject.core.entity.Technique</class>
+		<class>org.icatproject.core.entity.UserGroup</class>
+		<class>org.icatproject.core.entity.User</class>
+		<exclude-unlisted-classes />
```
RKrahl, line 65: Just a comment: I assume we'll need to update this list in any schema change that adds or removes an entity class in the future, such as in #366. I guess that will be very easily overlooked.

PR #395, opened 2026-03-27
Only rollback SessionManager.userTransaction if status is active

### LINKED ISSUE #394: SessionManager.logout attempts to rollback committed transactions
https://github.com/icatproject/icat.server/blob/9b0971b14d70c93e9b4269d2a88a710a148c76de/src/main/java/org/icatproject/core/manager/SessionManager.java#L151-L177

If calling logout on an expired session, we commit the transaction on line 156. At this point, `userTransaction` is completed, and does not need to/cannot be rolled back. If `logRequests` includes sessions, then we attempt `session.getUserName()` on line 161, which will throw an `IcatException` because the session is expired:

https://github.com/icatproject/icat.server/blob/9b0971b14d70c93e9b4269d2a88a710a148c76de/src/main/java/org/icatproject/core/entity/Session.java#L49-L53

This is caught, and a rollback attempted on line 167, but this throws an `IllegalStateException` because the transaction is no longer active as it was committed back on 156.

Note that this will only occur if `log.list` includes `SESSION`, so as an intermediate fix that can be changed in run.properties.

Propose that:
* In this case, do not throw for `getUserName`, as we are using it for sending a notification (and the fact that it has expired shouldn't matter as we're logging out anyway)
* Move the rollback to a `finally`, but only attempt it if the transaction is still active (nothing will happen if we have successfully committed, and in the case there's a non-IcatException before the commit we're also covered)
* Increase logging, as currently the error is returned to the client but does not appear in the icat.log, which complicated the debugging process
* Change the login and refresh methods to match for consistency

patrick-austin:
Closes #394

### DIFF
--- a/src/main/java/org/icatproject/core/manager/SessionManager.java
+++ b/src/main/java/org/icatproject/core/manager/SessionManager.java
@@ -12,6 +12,7 @@ import jakarta.json.stream.JsonGenerator;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.PersistenceContext;
 import jakarta.transaction.NotSupportedException;
+import jakarta.transaction.Status;
 import jakarta.transaction.SystemException;
 import jakarta.transaction.UserTransaction;
 
@@ -126,19 +127,26 @@ public class SessionManager {
 				}
 				return result;
 			} catch (Throwable e) {
-				userTransaction.rollback();
-				logger.trace("Transaction rolled back for login because of " + e.getClass() + " " + e.getMessage());
+				logger.trace("Attempt transaction roll back for login because of " + e.getClass() + " " + e.getMessage());
 				throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
 						"Unexpected DB response " + e.getClass() + " " + e.getMessage());
+			} finally {
+				if (userTransaction.getStatus() == Status.STATUS_ACTIVE) {
+					userTransaction.rollback();
+				}
 			}
 		} catch (IllegalStateException e) {
+			logger.error("IllegalStateException " + e.getMessage());
 			throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
 					"IllegalStateException " + e.getMessage());
 		} catch (SecurityException e) {
+			logger.error("SecurityException " + e.getMessage());
 			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SecurityException " + e.getMessage());
 		} catch (SystemException e) {
+			logger.error("SystemException " + e.getMessage());
 			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SystemException " + e.getMessage());
 		} catch (NotSupportedException e) {
+			logger.error("NotSupportedException " + e.getMessage());
 			throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
 					"NotSupportedException " + e.getMessage());
 		}
@@ -151,6 +159,7 @@ public class SessionManager {
 			try {
 				long startMillis = log ? System.currentTimeMillis() : 0;
 				Session session = getSession(sessionId);
+				String userName = session.getUserName(true);
 				sessionEntityManager.remove(session);
 				sessionEntityManager.flush();
 				userTransaction.commit();
@@ -158,14 +167,13 @@ public class SessionManager {
 				if (logRequests.contains(CallType.SESSION)) {
 					ByteArrayOutputStream baos = new ByteArrayOutputStream();
 					try (JsonGenerator gen = Json.createGenerator(baos).writeStartObject()) {
-						gen.write("userName", session.getUserName());
+						gen.write("userName", userName);
 						gen.writeEnd();
 					}
 					transmitter.processMessage("logout", ip, baos.toString(), startMillis);
 				}
 			} catch (IcatException e) {
-				userTransaction.rollback();
-				logger.trace("Transaction rolled back for logout because of " + e.getClass() + " " + e.getMessage());
+				logger.warn("Attempt transaction roll back for logout because of " + e.getClass() + " " + e.getMessage());
 				if (e.getType() == IcatExceptionType.SESSION) {
 					throw e;
 				} else {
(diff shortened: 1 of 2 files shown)


kevinphippsstfc (commented):
I've discussed this with Patrick and it looks OK to me but I think @ajkyffin should have a look over this as well.

ajkyffin (commented):
As discussed offline, I've tided up the exception handling so that fewer exceptions are caught, separated the 'logRequests' code from the database code, and improved the logging.

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-					transmitter.processMessage("login", ip, baos.toString(), startMillis);
-				}
-				return result;
-			} catch (Throwable e) {
-				logger.trace("Attempt transaction roll back for login because of " + e.getClass() + " " + e.getMessage());
-				throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
-						"Unexpected DB response " + e.getClass() + " " + e.getMessage());
+			} catch (PersistenceException e) {
+				logger.error("Database error", e);
+				throw new IcatException(IcatExceptionType.INTERNAL, "Database error: " + e.getMessage());
```
patrick-austin, line 124:
```suggestion
				throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that would be `PersistenceException` instead of `Database error`. Did you have a preference in the end?

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-		} catch (SystemException e) {
-			logger.error("SystemException " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SystemException " + e.getMessage());
-		} catch (NotSupportedException e) {
-			logger.error("NotSupportedException " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
-					"NotSupportedException " + e.getMessage());
+		} catch (HeuristicMixedException | HeuristicRollbackException | NotSupportedException | RollbackException | SystemException e) {
+			logger.error("Transaction error", e);
+			throw new IcatException(IcatExceptionType.INTERNAL, "Transaction error: " + e.getMessage());
```
patrick-austin, line 132:
```suggestion
			throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that wouldn't be `Transaction error`. Did you have a preference in the end?

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-				} else {
-					throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
-							e.getClass() + " " + e.getMessage());
-				}
-			} catch (Exception e) {
-				logger.warn("Attempt transaction roll back for logout because of " + e.getClass() + " " + e.getMessage());
-				throw new IcatException(IcatException.IcatExceptionType.INTERNAL, e.getClass() + " " + e.getMessage());
+			} catch (PersistenceException e) {
+				logger.error("Database error", e);
+				throw new IcatException(IcatExceptionType.INTERNAL, "Database error: " + e.getMessage());
```
patrick-austin, line 164:
```suggestion
				throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that would be `PersistenceException` instead of `Database error`. Did you have a preference in the end?

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SystemException" + e.getMessage());
-		} catch (NotSupportedException e) {
-			logger.error("NotSupportedException " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "NotSupportedException " + e.getMessage());
-		} catch (RuntimeException e) {
-			logger.error(e.getClass() + " " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, e.getClass() + " " + e.getMessage());
+		} catch (HeuristicMixedException | HeuristicRollbackException | NotSupportedException | RollbackException | SystemException e) {
+			logger.error("Transaction error", e);
+			throw new IcatException(IcatExceptionType.INTERNAL, "Transaction error: " + e.getMessage());
```
patrick-austin, line 172:
```suggestion
			throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that wouldn't be `Transaction error`. Did you have a preference in the end?

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-				} else {
-					throw new IcatException(IcatException.IcatExceptionType.INTERNAL,
-							e.getClass() + " " + e.getMessage());
-				}
-			} catch (Exception e) {
-				logger.warn("Attempt transaction roll back for refresh because of " + e.getClass() + " " + e.getMessage());
-				throw new IcatException(IcatException.IcatExceptionType.INTERNAL, e.getClass() + " " + e.getMessage());
+			} catch (PersistenceException e) {
+				logger.error("Database error", e);
+				throw new IcatException(IcatExceptionType.INTERNAL, "Database error: " + e.getMessage());
```
patrick-austin, line 201:
```suggestion
				throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that would be `PersistenceException` instead of `Database error`. Did you have a preference in the end?

src/main/java/org/icatproject/core/manager/SessionManager.java
```diff
@@ -93,167 +96,127 @@ public String getUserName(String sessionId) throws IcatException {
...
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SecurityException " + e.getMessage());
-		} catch (SystemException e) {
-			logger.error("SystemException " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "SystemException " + e.getMessage());
-		} catch (NotSupportedException e) {
-			logger.error("NotSupportedException " + e.getMessage());
-			throw new IcatException(IcatException.IcatExceptionType.INTERNAL, "NotSupportedException " + e.getMessage());
+		} catch (HeuristicMixedException | HeuristicRollbackException | NotSupportedException | RollbackException | SystemException e) {
+			logger.error("Transaction error", e);
+			throw new IcatException(IcatExceptionType.INTERNAL, "Transaction error: " + e.getMessage());
```
patrick-austin, line 209:
```suggestion
			throw new IcatException(IcatExceptionType.INTERNAL, e.toString());
```
We discussed the possibility of using `toString` instead, as this starts with the exception class name. Strictly speaking for this case that wouldn't be `Transaction error`. Did you have a preference in the end?

patrick-austin (commented):
@ajkyffin Just wanted to check about moving from `e.getMessage()` to `e.toString()`, or not, in the `new IcatException`s

